Issue #18, opened 2015-06-09
Does it support multi sites


amuntasim:
Using one installation, can we configure multiple sites

Bramjetten:
We implemented a multi-tenant version of Spina at denkwebsite.nl by decorating our models with default scopes.

deanperry:
I would be interested in this also. Could you possibly post the code you're using for denkwebsite.nl?

hanpanpan200:
I'd like to get some some help for multi-tenant too, could you please post the code for denkwebsite.nl?

mbajur:
Would be also interested in this. Meanwhile, i've found this gem extracted from discourse and i'm gonna try to use it:

https://github.com/discourse/rails_multisite

PR #30, opened 2015-06-14
Multi Site Support

deanperry:
Hey! I'm really liking Spina so far after finding it in the Ruby Weekly newsletter.

This is for #18. I'm still working on it but it's working very well so far.

In a new Rails app, add this to the routes

```ruby
constraints CustomDomainConstraint do
  mount Spina::Engine => '/'
end
```

and this to `lib/custom_domain_constraint.rb`

```ruby
class CustomDomainConstraint
  def self.matches? request
    Spina::Account.where(custom_domain: request.host).any? || Spina::Account.where(subdomain: request.subdomain).any?
  end
end
```

What do you think?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,14 @@
+GIT
+  remote: git://github.com/denkGroot/Spina-template.git
+  revision: 10503d6a28068f18ce9c2b0db76a8359a28c057a
+  specs:
+    spina-template (0.2.18)
+      bourbon (~> 3.2)
+      coffee-rails
+      fontcustom (~> 1.3)
+      jquery-fileupload-rails
+      neat (~> 1.5)
+
 PATH
   remote: .
   specs:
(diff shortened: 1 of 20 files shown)


deanperry:
To create a new account/site

```ruby
account = Spina::Account.create name: "New Site", theme: "default"
Spina::User.first.accounts << account
```

Bramjetten:
Do you think this should be a core component of Spina? For denkwebsite.nl we created a new model called Spina::Tenant. We added relations to all different models like pages and page parts. By adding default_scope -> { where(tenant_id: Tenant.current_id) } to these models we completely separate each tenant. Using a prepend_around_filter we set the current tenant:
    def current_tenant
      @current_tenant ||= Tenant.where('custom_domain = ? OR subdomain = ?', request.domain, request.subdomain).first
    end

deanperry:
That's completely up to you but if not then I'll just use it for my own use

PR #115, opened 2016-03-05
Allow configuring max page depth

lilfaf:
This was a major requirement for us so we thought it would be a nice addition to the project.
https://github.com/denkGroot/Spina-template/pull/9


lilfaf:
https://github.com/denkGroot/Spina-template/pull/9

Issue #114, opened 2016-03-03
Parent page input only showing top level parents


nicolasthy:
Hi !

When updating or creating a page, the `Parent page` input in the `Advanced` tab only shows the top level parents. Therefore when updating a sub child page, it sets its parent to the top level parent and you have to re-order your pages thereafter.

PR #134, opened 2016-04-21
Fix parent page selection #114

lilfaf:
Hi!
In addition to this fix, we probably should activate ancestry's depth caching and restrict presented pages to those that have a lower depth than the configured `max_page_depth`. Which would involve instructions for migrating because users would have to [rebuild the depth cache](https://github.com/stefankroes/ancestry#integrity-checking-and-restoration).

```ruby
def possible_parents
  parents = self.class.active.sorted.to_depth(Spina.config.max_page_depth - 1)
  new_record? ? parents : parents - subtree
end
```

### DIFF
--- a/app/models/spina/page.rb
+++ b/app/models/spina/page.rb
@@ -90,6 +90,11 @@ module Spina
       theme.view_templates.find { |template| template[:name] == view_template_name }
     end
 
+    def possible_parents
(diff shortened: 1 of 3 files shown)


harmdewit:
Having a max page depth could come in handy. I've included your addition of excluding the subtree from the select options in this commit 8e83f5f

PR #136, opened 2016-04-24
Fix urls in menu with non-standard backend_path

katafrakt:
This fixes issue similar to https://github.com/denkGroot/Spina/pull/132: When Spina is mounted not on `/`, links in navigation do not reflect it and lead to incorrect location (for example to `/test-page` instead of `/content/test-page` when `mount Spina::Engine => '/content/'`).

### DIFF
--- a/app/presenters/spina/pages/menu_presenter.rb
+++ b/app/presenters/spina/pages/menu_presenter.rb
@@ -60,8 +60,9 @@ module Spina
 
       def render_menu_item(menu_item, index, menu_items_length)
         content_tag(list_item_tag, class: menu_item_css(menu_item[0], index, menu_items_length)) do
+          path = File.join(Spina::Engine.routes.url_helpers.root_path, menu_item[0].materialized_path)
           buffer = ActiveSupport::SafeBuffer.new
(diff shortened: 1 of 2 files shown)


app/presenters/spina/pages/menu_presenter.rb
```diff
@@ -60,8 +60,9 @@ def render_menu_items(menu_items)
 
       def render_menu_item(menu_item, index, menu_items_length)
         content_tag(list_item_tag, class: menu_item_css(menu_item[0], index, menu_items_length)) do
+          path = File.join(Spina::Engine.routes.url_helpers.root_path, menu_item[0].materialized_path)
```
Bramjetten: Why use File for this?

katafrakt: It's most reliable method to join URL segments with avoiding duplicate slashes. Is there something with less misleading name?

Bramjetten: A simple .join would suffice in my opinion. We shouldn't have duplicated slashes anyway.

PR #147, opened 2016-05-28
Add property concern to keep page model thin and simple

arlyxiao:
@Bramjetten  Refactor page model to keep it simple and thin

### DIFF
--- a/app/models/concerns/spina/pages/property.rb
+++ b/app/models/concerns/spina/pages/property.rb
@@ -1,0 +1,66 @@
+module Spina
+  module Pages
+    module Property
+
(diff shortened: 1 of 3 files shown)


Bramjetten:
I don't see this as a particularly good refactoring. The code is simply spread across multiple files now. I'd rather have a bigger page.rb file.

arlyxiao:
@Bramjetten  Yep. this part is not important. The reason that all of the functions are very simple and close to the property/field. So I put it together.

It depends on you.

arlyxiao:
so close the PR?

PR #148, opened 2016-05-29
Refactor content in account model

arlyxiao:
@Bramjetten  Please review

### DIFF
--- a/app/models/spina/account.rb
+++ b/app/models/spina/account.rb
@@ -18,8 +18,7 @@ module Spina
     end
 
     def content(layout_part)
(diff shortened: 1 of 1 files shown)


Bramjetten:
This change raises an error when the record is not available.

arlyxiao:
@Bramjetten 
find_by will return first record or nil object
http://apidock.com/rails/v4.0.2/ActiveRecord/FinderMethods/find_by

It shouldn't have an exception.
Could you show the full error messages?

btw: which ruby version are u using?

PR #151, opened 2016-05-30
Remove rewrite_page

arlyxiao:
@Bramjetten  Please review

The function is not necessary. Because when the page can't be found out, it will raise an error shows the record not found

The method will catch it
rescue_from ActiveRecord::RecordNotFound, with: :render_404

### DIFF
--- a/app/controllers/spina/pages_controller.rb
+++ b/app/controllers/spina/pages_controller.rb
@@ -27,14 +27,13 @@ module Spina
       end
 
       def rewrite_page
-        unless page.present?
-          @rule = RewriteRule.find_by(old_path: "/" + params[:id])
-          redirect_to @rule.new_path, status: :moved_permanently if @rule.present?
+        if page.nil? && rule = RewriteRule.find_by(old_path: "/" + params[:id])
+          redirect_to rule.new_path, status: :moved_permanently
         end
       end
 
       def page_by_locale(locale)
-        Page.with_translations(locale).find_by!(materialized_path: spina_request_path)
+        Page.with_translations(locale).find_by(materialized_path: spina_request_path)
       end
 
       def page


Bramjetten:
What about RedirectRule?

arlyxiao:
@Bramjetten  The key point is that the function will not be executed.
Could u show an example that which kind of step will execute the function?

arlyxiao:
The code inside unless page.present? will not be executed

Bramjetten:
Why will it not be executed? It should be executed.

arlyxiao:
Because if the page can't be found out, it will have an error when execute the code 
Page.with_translations(locale).find_by!(materialized_path: spina_request_path)

and the code 
rescue_from ActiveRecord::RecordNotFound, with: :render_404
will catch the error

So it will not execute the code inside 
unless page.present?

Bramjetten:
Then that's a bug that should be fixed. We want URL's to be redirect to their previous URL when prossible.

arlyxiao:
Could u show an example?

Bramjetten:
Check out the after_save :rewrite_rule in page.rb.

arlyxiao:
Have checked that. And not clear about your requirement.

etc,
1, Created one new page and title = 'title 1'
2, Edit the page and set title = 'title 2'
3, Edit the page again, set the title = 'title 3'

The rewrite_rules table could have two records
title_1, title_2
title_2, title_3

So what's next? what's your requirement?
when visit /title_1, will it go to /title_2 ?
when visit /title_2, will go to /title_3 ?

Bramjetten:
That's right

arlyxiao:
@Bramjetten 
So finally, 
when visit /title_1, will go to /title_2, then go to /title_3

Bramjetten:
Yes

arlydemo:
@Bramjetten fixed it.
1, shouldn't use find_by! , should use find_by
2, a little refactor, it's not necessary to use the instance variable @rule in the local

PR #154, opened 2016-06-01
Refactor content in page model

arlyxiao:
@Bramjetten


Bramjetten:
When would this be necessary?

arlyxiao:
etc,
Add social function
login
register

PR #175, opened 2016-07-16
add linkedin and instagram inputs to settings

danmitchell-:
### Add Instagram and LinkedIn

Most businesses have more than just Facebook and Twitter so it's beneficial to have control over the other popular channels in the CMS too.


Bramjetten:
My proposal is to replace all social stuff with a single serialized field in the database. That way we can add as many services as we need.

DanBrooker:
How about something like `gem "rails-settings-cached"` which would allow arbitrary settings on account

PR #188, opened 2016-08-22
additional upgrade information in README

DanBrooker:
I'm working through updating to 0.10 and some extra information in README would have helped so I'm adding it.

Is there any data migration required when transitioning to 0.9 and the page translations? If so that should be mentioned as well.

Dropping the inquiries section should be mentioned somewhere, release notes perhaps? And a link to https://github.com/DigitalReflow/spina-inquiries would be nice

Also I have some plugins and noticed that the plugin examples haven't been updated since the design changes occurred.

I like SpinaCMS and want to see it improve and I think these suggestions could help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,12 @@ Spina::Theme.register do |theme|
 end
 ```
 
-Check out [config/initializers/themes/default.rb](https://github.com/denkGroot/Spina/blob/master/lib/generators/spina/templates/config/initializers/themes/default.rb) for an example.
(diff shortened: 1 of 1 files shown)


DanBrooker:
@Bramjetten thanks for adding this, here some other things that might be good to add

* Moving Carrierwave out of `initializers/spina.rb`
* Needs to still mention config.storage = :aws in `initializers/spina.rb`
* Adding a locale to `initializers/spina.rb` and running `rake spina:update_translations`

PR #203, opened 2016-11-29
Avoid conflicts with common authentication solutions

lilfaf:
In many cases, the main app might have its own authentication solution and we probably want to access the currently connected user from the pages (to display log in/out buttons for example).
Problem is that devise injects controllers helpers directly to `ActionController::Base` and therefore the `current_user` method is overriden by Spina.
Renaming the spina internal method to `current_spina_user` (which make sense in term of conventions) allows spina to play nicely with any authentication layer.


Bramjetten:
`current_user` is now `current_spina_user`

PR #216, opened 2017-01-30
Allow photo collections to be used in a structure

danmitchell-:
Currently you can't add a photo collection to a structure but this allows you to do that


Bramjetten:
I'll merge this, but I think we could refactor this to prevent some duplication.

PR #239, opened 2017-03-28
Add photo options

acapro:
when a photo in the wysiwyg editor is clicked, photo options appear
allowing for normal or wrapped class, and setting an alt tag


Bramjetten:
Safari currently does not support document.registerElement right now. Any way we can fix this?

acapro:
I added the polyfills from trix into another PR, supports Safari

Bramjetten:
I edited the modal slightly to bring it more in line with the rest of our layout. Still looking for a nice way to change the image class.

acapro:
Awsome!

Bramjetten:
Btw, is it necessary to require the entire source of Trix?

acapro:
Is it possible to overwrite class methods using $.extend? It would be a lot nicer that way, or maybe it  would make sense to branch out trix, modify it, and include the minified version of that. What do you think?

Bramjetten:
Closing for now, but will revisit when Trix offers better extendability.

daltonrenaldo:
@acapro this seems very useful, any update?

Bramjetten:
@daltonrenaldo This was partially implemented in a recently merged PR. It's possible to add alt-text and a URL to images now.

daltonrenaldo:
@Bramjetten you're correct. I decided to add to continue from that PR and add css classes. https://github.com/SpinaCMS/Spina/pull/509

PR #241, opened 2017-03-31
Remove legacy assets

danmitchell-:
This PR removes loads of dead weight :)

The last commit is personal preference; I like to have all assets together and not use the vendor directory but happy to revert. Thoughts?


Bramjetten:
Looks good, but I'm not sure about moving the vendor assets to the app folder. Any pros to do it this way?

danmitchell-:
The only pro is I like to see the assets all together but i've reverted that commit and rebased so you can merge :)

PR #257, opened 2017-06-05
Override 115% width for structure selects

simmerz:
This fixes the select so that it doesn't drop down to the line below the label


simmerz:
Anything preventing this being merged @Bramjetten ?

PR #258, opened 2017-06-08
Add `Spina::Link` type

xmjw:
I've only started using Spina in the last week, but I'm finding it a little tricky without this type to do what I want. I want to open an issue to suggest this, but I thought it would be nice to submit some code that suggests a possible solution?

This is to allow you to easily select a page on the site, with a title to generate a lind. This is useful for creating content blocks with (for example) an image, a line of text, and link button that leads to another
page.

I'm not totally happy with the way this is rendered in the UI, nor the `Spina::Link` model - but it's a start.

I'm also a bit at a loss as to how to test this.

### DIFF
--- a/PAGE_PARTS.md
+++ b/PAGE_PARTS.md
@@ -1,0 +1,127 @@
+# Page parts
+
+There are several code building blocks that can be used to create content in
+Spina.
+
+- `Spina::Line`
+- `Spina::Text`
+- `Spina::Photo`
+- `Spina::PhotoCollection`
+- `Spina::Structure`
+- `Spina::Option`
+
+## `Spina::Line`
+
+This is a single line of text. To add one to your template:
+
+    ::Spina::Theme.register do |theme|
+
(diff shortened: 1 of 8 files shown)


Bramjetten:
Hello and welcome! Thank you for your contribution. I like your proposal, it's a great start. There's a couple of things that could use improvement before we can use this in Spina.

- A form partial for structures
- A modal window similar to selecting photos for pages
- An extension for Trix to add links to pages in the same fashion

xmjw:
Sure, that sounds much better than the current UI side. The form partial for structures is a breeze, I'm still trying to pick apart the modal however, but I think that's a great plan.

I'm not really sure where to start with Trix, can you point me to a good resource?

simmerz:
@xmjw they haven't really developed the docs sufficiently yet. It's a case of reading the code and some of their examples. They do have a little documentation about how to add a "document" to a Trix editor instance.

PR #291, opened 2017-09-07
Amazon AWS s3

timbielawski:
To get spina to upload to S3 I added Fog/AWS  and had to update the default uploader

Is this needed? It is the only way it would work for me

 class DefaultStoreUploader < CarrierWave::Uploader::Base
storage :fog if Spina.config.storage == :s3

### DIFF
--- a/app/uploaders/spina/default_store_uploader.rb
+++ b/app/uploaders/spina/default_store_uploader.rb
@@ -1,7 +1,7 @@
 module Spina
   class DefaultStoreUploader < CarrierWave::Uploader::Base
 
-    storage :fog
+    storage :fog if Spina.config.storage == :s3
 
     def store_dir
       case ::Spina.config.storage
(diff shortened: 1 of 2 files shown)


Bramjetten:
No this is not needed. We prefer using a Carrierwave initializer file like this:

```ruby
# config/initializers/carrierwave.rb
CarrierWave.configure do |config|

  if Rails.env.production?
    # Use S3 if you want
    config.fog_credentials = {
      provider:               'AWS',
      region:                 Rails.application.secrets.aws_region,
      aws_access_key_id:      Rails.application.secrets.aws_access_key_id,
      aws_secret_access_key:  Rails.application.secrets.aws_secret_access_key
    }
    config.storage = :fog
    config.fog_directory  = Rails.application.secrets.aws_bucket
    config.fog_public     = true
    config.fog_attributes = { 'Cache-Control' => 'max-age=315576000' }
  else
    config.storage = :file
  end

  config.enable_processing = !Rails.env.test?
end
```

timbielawski:
yes thanks ! that works perfectly

PR #316, opened 2017-10-27
Added Polish (pl) localisation, remove duplicated entries for en.yml.

reelman:


reelman:
My `rake task` shows me:
```
28 tests, 41 assertions, 0 failures, 0 errors, 0 skips
```
not sure why CI failed.

Bramjetten:
There seems to be a timing issue in our tests which causes CI to fail sometimes. Often it's enough to simply rebuild.

PR #322, opened 2017-11-15
add fog_provider to carrierwave template

andywenk:
when not including the config.fog_provider = 'fog/aws' starting the Rails application will result in the error

.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activesupport-5.1.4/lib/active_support/dependencies.rb:292:in `require': cannot load such file -- fog (LoadError)

### DIFF
--- a/lib/generators/spina/templates/config/initializers/carrierwave.rb
+++ b/lib/generators/spina/templates/config/initializers/carrierwave.rb
@@ -2,6 +2,7 @@ CarrierWave.configure do |config|
   config.storage = :file
 
   # Use S3 if you want
+  # config.fog_provider = 'fog/aws'
   # config.fog_credentials = {
   #   provider:               'AWS',
   #   region:                 '',


andywenk:
oh I am not sure why CI fails. This has nothing to do with my change I hope:

```
Failure:
Spina::PagesControllerTest#test_visit_homepage [/home/ubuntu/Spina/test/functional/spina/pages_controller_test.rb:12]
Minitest::Assertion: Expected response to be a <2XX: success>, but was a <404: Not Found>
```

danmitchell-:
Is this for when you use the fog-aws gem instead of the generic fog gem?

andywenk:
yes I am using the `fog-aws` gem. It is not recommended to use the `fog` gem if you do know the provider.

danmitchell-:
Nice work 👍

Bramjetten:
Nice find, merging it. Small aside: we will try switching to ActiveStorage soon. 

CI is failing because of some weird issue in our tests, rebuilding it fixes it.

andywenk:
@Bramjetten oh cool - ActiveStorage is nice. Thanks for the merge. It would be cool to have section in the README about how to use S3. If I find a time slot I will send another PR.

PR #326, opened 2017-11-17
Adding ro.yml

TrueSoft:
I added a locale file for Romanian language


Bramjetten:
The yaml file begins with `en`.

PR #356, opened 2018-04-21
Support for Spina::Resource

Bramjetten:
Often we'd like to manage things in Spina besides pages. Examples are a portfolio, simple blogposts or (in our case) a list of breweries. Usually we want those to have a URL and page parts, just like pages. So we end up recreating a lot of functionality in a new plugin.

In order to make this a lot easier we'd like to introduce a new model called `Spina::Resource`. 

It's actually quite simple. This model stores a few attributes regarding our resource: `name`, `label`, `view_template`, `order_by`, `parent_page`.

## Example
Say we want to manage a list of breweries (which we do).

We can create a resource like this:
```ruby
Spina::Resource.create(
  name: "breweries", 
  label: "Breweries", 
  order_by: "title", 
  view_template: "brewery"
)
```
This immediately creates a new item in our website navigation:

schermafbeelding 2018-04-21 om 13 12 59

When I click on the add button, I'm greeted by a new page form:

schermafbeelding 2018-04-21 om 13 14 08

This is a page just like any other in Spina, including page parts, URL rewrites and ancestry. The exception is that these pages only show up inside the breweries section of Spina because these pages belong to our newly created resource. 

If we'd like to get our breweries for our frontend we can simply do this:

```ruby
Spina::Resource.find_by(name: "breweries").pages
```

A nice helper could be:

```ruby
def breweries
  @breweries ||= Spina::Resource.where(name: "breweries").first_or_create.pages
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,13 +25,22 @@ The installer will help you setup your first user.
 
 Then start `rails s` and access Spina at `/admin`.
 
-## Upgrading from 0.12 to 1.0.0.alpha
+## Upgrading from 0.13 to 1.0.0.alpha
 
-Going from 0.12 to 1.0 will introduce a couple of changes. Globalize is replaced by Mobility and later down the road CarrierWave will be replaced by Rails' ActiveStorage. 
+Going from 0.13 to 1.0 will introduce a couple of changes. Globalize is replaced by Mobility and later down the road CarrierWave will be replaced by Rails' ActiveStorage. 
 
 Switching to Mobility is fairly straightfoward. 
 - Run `rails g spina:install` to add the `mobility.rb` initializer.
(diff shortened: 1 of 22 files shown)


danmitchell-:
This is a great idea. I’m doing this exact thing for a portfolio at the moment 👍

Bramjetten:
One downside I see is that in order to get all of your 'regular' pages, I had to introduce two new scopes on the Page model: `regular_pages` and `resource_pages`.

You only have to use it if you're using `Spina::Resource` though, so perhaps it's not a big deal.

Bramjetten:
Also, a great addition would be to create a `Spina::PagePart` to select pages or resources.

Issue #348, opened 2018-03-12
Materialized path does not generate correct paths for all locales


danmitchell-:
When a regional locale is specified i.e. `zn-CN` the path generated is incorrect.

Issue #220, opened 2017-02-22
Localized homepages - materialized path problem


jhovad:
Materialized path for localized homepage is generated like "/de/". It have to be without ending slash. I suggest to update localized_materialized_path in Spina/app/models/spina/page.rb 

```ruby
def localized_materialized_path
  if I18n.locale == I18n.default_locale
    generate_materialized_path
  else
#my edit:
    if root? && name == 'homepage'
      generate_materialized_path.prepend("/#{I18n.locale}")
    else
      generate_materialized_path.prepend("/#{I18n.locale}/")
    end
#end of edit.
  end
end
```

PR #359, opened 2018-04-27
Feature/url title improve

dminchev:
This change will solve the problem with bad materialized_path generation for pages with title in different than latin chars.

I'm using babosa gem with currently supports the following languages:
- bulgarian
- cyrillic
- danish
- german
- greek
- latin
- macedonian
- norwegian
- romanian
- russian
- serbian
- spanish
- swedish
- ukrainian
- vietnamese

In order to use activate some of the languages different from latin you have to specify in initializer similar to locales:

```ruby
Spina.configure do |config|
  config.transliterations = %i(latin bulgarian german)
end
``` 

Also I've created a fallback behavior if the title parameterization fails to build the slug based on the page id 

```ruby
"page-#{id}"
```

Here is reference to issues with materialized_path #348 #220 

Hope you find this useful!


Bramjetten:
What does this offer that `parameterize` does not?

dminchev:
Title in languages with different from latin chars don't generate proper path

```
[1] pry(main)> "Тест".parameterize
=> ""
[2] pry(main)> "Test".parameterize
=> "test"
```
with this changes the string will be translated to proper url friendly latin chars or will fallback to page-id

Bramjetten:
Do we have to define all transliterations in a config file? Can't we just add all available transliterations?

dminchev:
I don't think it will be a problem to add by default all, however I've checked similar configs for friendly-id and non latin title and they follow this pattern. 
My personal opinion is to list just the ones you need.

Bramjetten:
If we merge this, the default should be that no changes are required for existing projects. Anyone who wants to add transliterations could do so using this config. Perhaps we could add some comments in Spina's config file as an example.

dminchev:
Fair enough, I think that running install would add the default latin from the template. However this is additional action to the devs. I'll add default latin in the config so it won't require any other actions for the existing projects.

```
rails g spina:install
```

Will add it to the readme.

dankmitchell:
@dminchev I think this is a great improvement, can you replace to fix conflicts and add a note into he readme.

Bramjetten:
Is this still a wanted feature?

Bramjetten:
I'm going to assume this is still a wanted feature. Since it's quite an old PR it'll need some editing. @dminchev cool if I pick this up?

Bramjetten (approved):

Bramjetten:
I'm sorry for taking so long, but it's merged now! ✌️

dminchev:
> I'm going to assume this is still a wanted feature. Since it's quite an old PR it'll need some editing. @dminchev cool if I pick this up?

Yeah sure, thanks!

PR #360, opened 2018-04-27
feature/facebook pixel

dminchev:
Small feature allowing users to set Facebook pixel from the administration (as google analytics)


Bramjetten:
We're moving away from these hardcoded preferences in favor of general settings.

PR #361, opened 2018-04-30
Feature/navigation label translations

dminchev:
Added navigation group translations. This will allow users to set different label for each locale in navigations groups. 
This is useful if you have many navigations in your site (footer navigations groups, example: http://take.ms/jRfTq)

After updating you have to run 
```
rails spina:install:migrations
rails db:migrate
```

After successful installation there will be similar to pages edit switch with option to switch locale and enter label for each locale (http://take.ms/urijT).

Hope you find this useful.

### DIFF
--- a/app/controllers/spina/admin/admin_controller.rb
+++ b/app/controllers/spina/admin/admin_controller.rb
@@ -23,6 +23,10 @@ module Spina
         render status: 401 unless current_spina_user.admin?
       end
 
+      def set_locale
+        @locale = params[:locale] || I18n.default_locale
+      end
+
     end
   end
 end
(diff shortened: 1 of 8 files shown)


Bramjetten:
I'm not sure if translated labels are needed for `Spina::Navigation`. Spina's backend was designed to use with 1 language. Adding translations for these kinds of labels means we'll have to add them in a lot more places as well and I'm not sure if it's worth the added complexity.

dminchev:
The idea here is not to translate the admin labels, but to allow users to have many navs (like footer navs) with labels for each language so you can directly display the nav label on the frontend without need to add it to locales. Same as you create different links display titles for each languge.

Bramjetten:
This feature was changed dramatically in the current release of Spina. Closing this.

PR #380, opened 2018-07-03
Fix: Missing remote attribute in media picker form

joeswann:
Without this attribute the form submits the page rather than sending an ajax request

### DIFF
--- a/app/views/spina/admin/media_picker/_modal.html.haml
+++ b/app/views/spina/admin/media_picker/_modal.html.haml
@@ -4,7 +4,7 @@
       .item.item-small.item-uploader.new-image-form
         = render partial: 'spina/admin/images/form'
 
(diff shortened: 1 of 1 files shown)


Bramjetten:
form_with generates remote forms by default. Local requests are the exception here.

joeswann:
Cool, thought I might be missing something.

This fixed an issue I was having with integration into an existing app, presumably due to an app or engine level override previously made to remote forms.

Note for context: I made a  #support channel comment on the Spina CMS slack about this https://spinacms.slack.com/archives/C0TR7K0J0/p1530586838000128

dankmitchell:
Is this still an issue @joeswann. Ive never seen it the error..

lukaVarga:
@dankmitchell I've just upgraded to Spina 1.0.3 from 0.12 and I am also having the exact same issue as @joeswann and his PR seems to fix it. Are there any plans to merge this PR to master?

Issue #385, opened 2018-07-15
Spina::Admin::MediaPickerController#select is missing a template for this request format and variant.


abunsen:
Whenever I try to pick an image from the media picker, I get a 406 error, bubbling up from rails.
![screen shot 2018-07-15 at 6 28 48 pm](https://user-images.githubusercontent.com/93429/42739025-f9d7ca58-885c-11e8-8d09-53d964809732.png)

```I, [2018-07-15T22:27:47.993764 #17332]  INFO -- : [552f1f6b-b0b8-4bba-b250-47c3c767186e] Started POST "/guides/admin/media_picker" for 75.74.122.39 at 2018-07-15 22:27:47 +0000
I, [2018-07-15T22:27:47.994730 #17332]  INFO -- : [552f1f6b-b0b8-4bba-b250-47c3c767186e] Processing by Spina::Admin::MediaPickerController#select as HTML
I, [2018-07-15T22:27:47.994884 #17332]  INFO -- : [552f1f6b-b0b8-4bba-b250-47c3c767186e]   Parameters: {"utf8"=>"✓", "authenticity_token"=>"tD45cJXElenjy71CPcefty6CrspJdTtND2+8csKVKsA3iy8xUosOwuT+F5cMzvkG0ml+ulvrMX6eoyVCoGriuw==", "hidden_field_id"=>"media_picker_47031524642720", "trix_editor_id"=>"", "image_id"=>"2", "image_ids"=>["2"]}
D, [2018-07-15T22:27:47.998177 #17332] DEBUG -- : [552f1f6b-b0b8-4bba-b250-47c3c767186e]   Spina::User Load (0.7ms)  SELECT  "spina_users".* FROM "spina_users" WHERE "spina_users"."id" = $1 ORDER BY "spina_users"."id" ASC LIMIT $2  [["id", 1], ["LIMIT", 1]]
D, [2018-07-15T22:27:47.999481 #17332] DEBUG -- : [552f1f6b-b0b8-4bba-b250-47c3c767186e]   Spina::Image Load (0.6ms)  SELECT  "spina_images".* FROM "spina_images" WHERE "spina_images"."id" = $1 LIMIT $2  [["id", 2], ["LIMIT", 1]]
I, [2018-07-15T22:27:48.000329 #17332]  INFO -- : [552f1f6b-b0b8-4bba-b250-47c3c767186e] Completed 406 Not Acceptable in 5ms (ActiveRecord: 1.3ms)
[...]

PR #386, opened 2018-07-16
Local false, bug fix for modal form

abunsen:
Hey! Thanks for all of the hard work on Spina. We ran into a bug with Spina (#385) from the mater branch with rails 5.2. So I hopped in the code and forced the SJR to happen by adding `local: false` to the modal.

Not sure if this is an issue or not for anyone else, but it's here if you want to merge it.

### DIFF
--- a/app/views/spina/admin/media_picker/_modal.html.haml
+++ b/app/views/spina/admin/media_picker/_modal.html.haml
@@ -4,7 +4,7 @@
       .item.item-small.item-uploader.new-image-form
         = render partial: 'spina/admin/images/form'
 
-      = form_with url: admin_media_picker_path, class: 'gallery-prepend-image' do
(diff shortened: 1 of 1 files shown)


Bramjetten:
Remote forms are the default for the form_with helper. I’m not sure why adding local: false would help. Are you overriding something elsewhere?

abunsen:
>Remote forms are the default for the form_with helper. I’m not sure why adding local: false would help. Are you overriding something elsewhere?

@Bramjetten no we are using an existing rails 5.2 app. I’m fairly sure form_with is not being overridden anywhere!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dankmitchell:
Closing in favour of https://github.com/SpinaCMS/Spina/pull/380 if still an issue

Issue #383, opened 2018-07-11
Support SVGs


podruh:
Hey, Spina should support uploading SVGs to Photo model.
We had to override `PhotoUploader` with our own and then the same with `uploads.coffee`, but Rails decided to use vanilla version of the script in production mode. So we ended with copying Spina gem to vendor and edit the script there. 
If there is another solution, I'm open for it.

Bramjetten:
In the latest version of Spina we switched to the new `Spina::Image` model which uses ActiveStorage. I haven't tested SVG's yet, but I suppose it should work like expected.

podruh:
I have checked the 'uploads.coffee' an master and it seems the same as the one in 0.12.0.
BTW: we are stuck on 0.12.0, because of Rails 5.1 and other reasons

dankmitchell:
Stack trace 

```
ActionView::Template::Error (ActiveStorage::InvariableError):
2018-08-29T11:41:54.947144+00:00 app[web.1]: F, [2018-08-29T11:41:54.947087 #4] FATAL -- : [c18b250f-4a12-4a30-8578-445a08864720]      7:     = radio_button_tag :image_id, image.id
2018-08-29T11:41:54.947145+00:00 app[web.1]: [c18b250f-4a12-4a30-8578-445a08864720]      8:     = check_box_tag 'image_ids[]', image.id, image.id.to_s.try(:in?, params[:selected_ids].to_a)
2018-08-29T11:41:54.947147+00:00 app[web.1]: [c18b250f-4a12-4a30-8578-445a08864720]      9:
2018-08-29T11:41:54.947148+00:00 app[web.1]: [c18b250f-4a12-4a30-8578-445a08864720]     10:   = image_tag main_app.url_for(image.variant(resize: '300x300^', crop: "300x300+0+0"))
```

Bramjetten:
I suppose due to SVG's nature, it's hard to resize/crop those images.

dankmitchell:
Solution for this is to not call `variant` on SVG file types.

PR #395, opened 2018-10-22
Add SVG content type check

podruh:
This originates from #383, originally from 0.12 issue, but is easily fixable in 1.0.x
It just check if content type includes `svg` string and returns file without variant if matches.


dankmitchell:
Sorry for the delay on this one @podruh, would you mind resolving the conflict so we can get this over the line?

podruh:
@dankmitchell Sorry for my delay. I removed my initial implementation in `images_helper.rb` because at the time I didn't know what I was really doing.

PR #409, opened 2019-01-19
Store page content as JSON

Bramjetten:
I want to greatly speed up websites built with Spina. Right now pages contain many parts. Many of these parts have translations. Because of the way Spina works this means having a lot of queries to fetch the required records when building a simple page with a dozen different page parts. The only way to speed up your website is to use Rails' `cache` method heavily. 

I think we can improve this process by storing the content of an entire page as JSON. This way we only have to run two SQL queries: one for the page and another for the page's translation.

This PR is a rough sketch. I'm not sure if we should use the `content` method for this.


Bramjetten:
Perhaps we should just switch to JSON entirely for page content. Not just when rendering a page, but also when creating or editing one in Spina.

DanBrooker:
I think this has merit, it's unlikely most people would be querying pages by page part. 
Also I'd be interested in a versioning system, i.e. being able to rollback or have an unpublished edit.

PR #414, opened 2019-03-22
Adding missing images translations

ErvalhouS:


config/locales/pt-BR.yml
```diff
@@ -125,6 +140,7 @@ pt-BR:
 
     media_library:
       photos: Imagens
```
dankmitchell: Is `photos` used anymore?

PR #415, opened 2019-03-25
Add documentation for Spina::Resource and add resources to theme config

Bramjetten:
I added much needed documentation for `Spina::Resource`. 

This PR also enables creation of resources using your theme config.


.ruby-version
```diff
@@ -1 +1 @@
-2.5.0
+2.5.1
```
dankmitchell: Let’s go to 2.6.1 whilst we’re making this change

simmerz: We should probably leverage travis' multi-ruby-version matrix capabilities and test against all Rails 5.2 supported versions of Ruby

PR #443, opened 2019-06-16
Update README.md

cprodhomme:


cprodhomme:
the color syntax work better like this

PR #467, opened 2019-09-22
Fix admin#edit images (missing helper)

abunsen:
### Context
When trying to view a recently uploaded image, I encountered this error:

Screen Shot 2019-09-21 at 11 11 01 PM

Was encountering this error, I traced it to the `Spina::ImagesHelper` not being autoloaded, so I manually load it from the autoloaded module `Spina::Admin::PagesHelper`.

### Guidance to review
Would love to get this merged because I'm trying to take a project live shortly.


abunsen:
For anyone else: I had `config.action_controller.include_all_helpers` set to `false` - this was the reason I was getting the above error.

PR #509, opened 2020-01-04
Insert image css class

daltonrenaldo:
### Context
When Adding an image, to a page,  it would be nice to be able to add specific classes to the image so that we can style and position it on the frontend. 

This PR extends the recent alt text, and link url feature to let the user select positioning classes to add to the image.

ezgif com-video-to-gif


dankmitchell:
I think the code looks fine but I'm not convinced on the radio buttons. Perhaps a `select` element would fit better with the UI?

Bramjetten:
The UI needs work. Also, are these classes user definable? Should we make them configurable? How can you see which class is selected in Trix? I believe this will require extensions in Trix to pull off really well.

daltonrenaldo:
I'll try to come up with something better soon.

PR #515, opened 2020-01-11
Coffee to js

dankmitchell:
### Context
ES6 has removed any motivation to continue using Coffeescript. We shouldn't maintain it.

Convert all `.coffee` to `.js` using  [CoffeeScript Cli](https://coffeescript.org/#cli) i.e. `coffee -wc .`

### Guidance to review
Everything  should work as before 🤔


Bramjetten:
I agree that we should move to ES6, but what's the added benefit of using this CLI to generate javascript that's a bit obfuscated? 

I'd propose we rewrite our JavaScript as Stimulus controllers where possible. That way we can move to ES6 ánd have a way better structure to deal with js.

PR #529, opened 2020-02-25
Redesign Page form and media picker

Bramjetten:
I've redesigned the page form and media picker in Spina. Everything's a bit cleaner now. There's more space to write now.

The media picker is now built with a couple of simple Stimulus controllers instead of jQuery. CSS was also improved.

Schermafbeelding 2020-02-25 om 15 09 05

Schermafbeelding 2020-02-25 om 15 08 34


Bramjetten:
This will be merged in #543

PR #533, opened 2020-03-25
Structure returns structure_items by default

Bramjetten:
* Structures now return `structure_items` directly instead of `self`
* Added more structure to CHANGELOG.md
* Added UPGRADING.md

### Structures in views
When calling `content` on a `Structure` page part Spina will now return the underlying items instead of the structure itself. That means you need to change all views using a structure. 

Before:
```ruby
content(:a_structure).structure_items.sorted_by_structure.each do |item|
  # ...
end
```

After:
```ruby
content(:a_structure).each do |item|
  # ...
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,30 +1,44 @@
 # Spina CMS Changelog
 
-## 1.1
+## Unreleased
 
-### 1.1.3 (December 25, 2019)
+### Changed
+* Structures now return `structure_items` directly instead of `self`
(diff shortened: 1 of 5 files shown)


Bramjetten:
@simmerz @dankmitchell Requested mostly to review the way I changed the changelog and upgrading files. Like Tom suggested, I think we could benefit from a Dangerfile as well to enforce these things.

Bramjetten:
Closing this in favor of Spina v2.

PR #541, opened 2020-04-27
Allow SVGs to be displayed

simmerz:
### Context

SVGs don't display in the media library or in image pickers as they are not variable. This patch fixes that.

As ActiveStorage > 5.2.1 forces SVGs to be downloaded as attachments due to XSS risks (https://github.com/rails/rails/issues/34665#issuecomment-445888009) we need to tell users to remove `image/svg+xml` from the sanitiser. Example provided [here](https://github.com/Dreamersoul/administrate-field-active_storage/issues/36#issuecomment-608446819).

I've written this into the pull request, but it will probably need adding to the guides.

### Guidance to review

Upload an SVG to the media library. I'm working locally, so this was tested using local `disk` storage. Then add the initializer and restart.

### DIFF
--- a/app/helpers/spina/images_helper.rb
+++ b/app/helpers/spina/images_helper.rb
@@ -8,6 +8,14 @@ module Spina
       if image.attached? && image.variable?
         variant = image.variant(options)
         main_app.rails_blob_representation_path(variant.blob.signed_id, variant.variation.key, variant.blob.filename)
+      elsif image.image?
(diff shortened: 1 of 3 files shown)


simmerz:
> Personally not a fan of letting end users deal with SVG’s

Neither am I, but we've been explicit about the reasons for needing it. I'm not a fan of using Spina as a UGC platform anyway, but then I'm not all users :)

PR #582, opened 2020-07-07
Rails to 6.0.3.2

Bramjetten:
For some reason we ended up with `gem 'rails', github: 'rails/rails'` on our master branch, resulting in a PR from dependabot for every commit to the Rails repository. Changed it to the regular Rubygems releases. (6.0.3.2 atm)


Bramjetten:
Never mind, we were already targeting some stuff for Rails 6.1 in our testing. I'll wait for Rails 6.1 first.

PR #584, opened 2020-07-08
Trix toolbar sticky - Fixes #579

### LINKED ISSUE #579: Trix editor blocks should keep the toolbar visible at all times
When writing long-form text content (such as blogs), the toolbar goes out of sight and so there's a lot of scrolling up and down to add links, images etc.

A quick and dirty (read hacky) solution seems to be `position: sticky; top: 117px;` but that assumes the header bar is always the same size across plugins and pages, which we shouldn't necessarily assume.

Bramjetten:
Make the toolbar in trix sticky. Adding the correct `top` to the toolbar happens in javascript, just like adding `paddingTop` for `section#main`.

### DIFF
--- a/app/assets/javascripts/spina/admin/scaffold.coffee
+++ b/app/assets/javascripts/spina/admin/scaffold.coffee
@@ -65,6 +65,7 @@ ready = ->
   if header = document.getElementById('header')
     headerHeight = header.getBoundingClientRect().height
     $('section#main').css({paddingTop: headerHeight})
+    $('section#main trix-toolbar').css({top: headerHeight + 15})
 
   # Login wrapper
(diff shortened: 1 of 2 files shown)


Bramjetten:
@simmerz Rewriting the entire page with better CSS would be better, but requires many changes. I suggest we add this "hack" to make Trix easier to work with now, and rewrite the CSS properly later.

simmerz:
@Bramjetten did you manage to solve the editor flowing over the toolbar yet?

Bramjetten:
> @Bramjetten did you manage to solve the editor flowing over the toolbar yet?

Not yet. Note to self: fix z-index and background color of toolbar.

PR #602, opened 2020-07-23
Resize images embedded in Trix

Bramjetten:
New configuration option:
```ruby
Spina.config.embedded_image_size = "2000x2000>"
```

Images selected and inserted in Trix are now resized by default. It makes no sense to always insert the original file URL, because it can potentially be way too huge for any content field. Spina resizes images to fit 2000x2000px by default. This is configurable, so you can optimize this for your own project. (I've changed it to 1200x1200 for example)

This only applies to images embedded in Trix.


jeroenson:
👊

PR #612, opened 2020-08-23
Spina::Parts::MultiLine

daltonrenaldo:
### Context
I find myself having to add this in every spina project that I do, I thought maybe it is generic enough to be part of spina. The use case is something you want text longer than `line`, but also want to limit the field to only rich-text (so the editor doesn't work either). This can be used for things like blog excerpts, quotes, etc... just whenever `line` and `text` won't do.

Creating a new part, Spina::Parts::Multiline, which is similar to line, but uses a textarea instead of an input field


simmerz:
Sounds reasonable to me, though personally I always use the spina-blog engine for blog posts.

I guess you'd use `simple_format()` around the content?

@bramjetten you happy to merge this in?

Bramjetten (approved):

Bramjetten:
Approved. I like it. Also think the name aptly describes it’s use case.

daltonrenaldo:
sweet, I'll update changelog and docs.

PR #622, opened 2020-10-08
Strip HTML & truncate repeater text

markmead:
### Context

Cleans up repeater text in the admin to truncate if too long, and to strip HTML if a text part is the first field

The reason for the `display: flex` change can be seen [on CodePen](https://codepen.io/markmead/pen/dyXPOpb)


Bramjetten (approved):
Thanks Mark! 🙏

PR #624, opened 2020-10-23
Make optional loading of decorators

jalkoby:
In rails 6 spina loaded decorators twice which was not ok especially if there're constants. also it was unexpected that decorators are loaded automatically - so I add ability to turn off


Bramjetten:
Why were decorators loaded twice?

jalkoby:
Well it's not clear for me why. But it definitely relates to bootsnap. the library override "Kernel#load" to provide caching mechanism and tries to load file a few times as you can see:

```ruby
alias_method(:load_without_bootsnap, :load)
  def load(path, wrap = false)
    if (resolved = Bootsnap::LoadPathCache.load_path_cache.find(path))
      return load_without_bootsnap(resolved, wrap)
    end

    # load also allows relative paths from pwd even when not in $:
    if File.exist?(relative = File.expand_path(path))
      return load_without_bootsnap(relative, wrap)
    end

    raise(Bootsnap::LoadPathCache::CoreExt.make_load_error(path))
  rescue LoadError => e
    e.instance_variable_set(Bootsnap::LoadPathCache::ERROR_TAG_IVAR, true)
    raise(e)
  rescue Bootsnap::LoadPathCache::ReturnFalse
    false
  rescue Bootsnap::LoadPathCache::FallbackScan
    load_without_bootsnap(path, wrap)
  end
```

jalkoby:
when I've replace `load` with `require_dependency` everything loaded as expected

jalkoby:
any updates?

josephschito:
I have the same problem:

`/Users/joseph/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3/lib/active_record/associations.rb:1940: warning: already initialized constant Spina::Account::HABTM_SpinaUsers
/Users/joseph/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/activerecord-6.1.3/lib/active_record/associations.rb:1940: warning: previous definition of HABTM_SpinaUsers was here`

Bramjetten:
@josephschito When are you seeing this? Can you reproduce this in a new Rails app?

josephschito:
@Bramjetten 

Try to add `app/decorators/spina/account_decorator.rb` file with the following content and then restart/reload! server/console:

```
require_dependency 'spina/account'

module Spina
  module AccountDecorator
    def self.prepended(base)
      base.class_eval do
        has_and_belongs_to_many(
          :spina_users,
          class_name: '::Spina::User',
          join_table: 'spina_accounts_users',
          association_foreign_key: 'spina_user_id',
          foreign_key: 'spina_account_id'
        ) 
      end
    end
    
    ::Spina::Account.prepend self
  end  
end
```

Bramjetten (approved):

Bramjetten:
Sorry for taking so long getting this merged!

PR #630, opened 2020-11-22
Spina::Parts::Datetime to add datepicker to theme

daltonrenaldo:
### Context
Sometimes you may want to add a date to your themes. This can be used for a post publication date. It is also helpful to have a Datetime object to manipulate on the frontend (for whatever reason). 

A new part type that allows me to select a date and time using a datepicker.

### Guidance to review

Screen-Recording-2020-11-22-at-12 12 50-PM


Bramjetten:
I’m open to the idea of supporting date/time inputs, but I’m not a fan of datepickers. I prefer having input fields with something like https://nosir.github.io/cleave.js/

Bramjetten:
The current version of Spina includes lots of UI changes this PR doesn't include yet. My suggestion is to add this feature as a plugin to Spina instead of adding it to Spina's core.

Issue #642, opened 2021-01-03
Rails 6 ActiveStorage switch from mini_magick to image_processing breaks image upload


wakproductions:
Rails changed a dependency that breaks Spina if you're using Rails 6. They changed the image processing gem within ActiveStorage. 

https://edgeguides.rubyonrails.org/active_storage_overview.html
http://railsdiff.org/5.2.4.4/6.0.0
[A nice overview of the changes in a blog post](https://bloggie.io/@kinopyo/upgrade-guide-active-storage-in-rails-6)

I would like to make a pull request to fix this, but not sure of the architectural approach you'd like to handle it. I'm considering two options:

1) Make a change to `Spina` that switches dependency from `mini_magick` to `image_processing`, and make the appropriate changes to the resizing method calls. The disadvantage of this is that Spina will no longer be compatible with Rails 5.

2) Try to abstract the image processing routines so that both `mini_magick` and `image_processing` are kept as dependencies, and somewhere in the background it will change the image handling appropriately depending on whether you are using Rails 5 or 6. This approach is going to take a lot more work, but one of the benefits is that by abstracting the image processor to a generic internal interface, we might be able to support other image hosting platforms like CarrierWave (yes, I saw there's another open issue requesting that).

If you give me some guidance on what you think the best approach is, I'd be happy to try taking that approach in a PR.

Bramjetten:
This is already fixed in the latest 2.0.0.alpha :)

wakproductions:
I'm working off of the master branch, which I think is more current than v2.0.0.alpha.... looks like it's still using `mini_magick` as a dependency and making the old style method calls. I got this going on when trying to load an image:

![image](https://user-images.githubusercontent.com/1912237/103488556-8f29f400-4ddb-11eb-88f8-c707185c59d9.png)

wakproductions:
I am able to resolve this issue by adding `  s.add_dependency 'image_processing'` to the `spina.gemspec` file. But, I'm having another issue when using the image upload form on the Page editor. The ActiveStorage Direct Upload feature seems to be creating the blobs alright, but the `ImagesController` fails because it doesn't recognize the request as being an AJAX request and raises a template not found error because there's no `create.haml` file to render, only a `create.js.erb`, which it doesn't reach because it doesn't think this is AJAX.
![image](https://user-images.githubusercontent.com/1912237/103498044-6c1a3700-4e11-11eb-93b2-409d94a7baff.png)

[...]

PR #643, opened 2021-01-04
Fix broken file upload return routes

wakproductions:
### Context

* Image and document uploads do not work correctly due to the forms not using XHR.
* ~~ActiveStorage is unable to work in Rails 6 due to missing dependency on `image_processing` gem~~

See #642 for further details

* Add `local: false` to the `form_with` forms
* ~~Add the `image_processing` gem~~

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,6 @@ PATH
       breadcrumbs_on_rails
       coffee-rails
       haml-rails
-      image_processing
       jquery-rails
       jsonb_accessor (>= 1.0.0)
       kaminari
@@ -126,9 +125,6 @@ GEM
       ruby_parser (~> 3.5)
     i18n (1.8.5)
       concurrent-ruby (~> 1.0)
-    image_processing (1.12.1)
-      mini_magick (>= 4.9.5, < 5)
-      ruby-vips (>= 2.0.17, < 3)
     jquery-rails (4.4.0)
       rails-dom-testing (>= 1, < 3)
       railties (>= 4.2.0)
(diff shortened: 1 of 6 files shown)


Bramjetten:
The `image_processing` gem is part every new Rails app and can be uncommented in the main app's Gemfile. Setting up ActiveStorage is a prerequisite for installing Spina so we prefer not adding it to Spina's gemspec.

As for `form_with` generating remote forms: this should already be the default in new Rails apps. You can make sure Rails generates remote forms by adding this config option:
`Rails.application.config.action_view.form_with_generates_remote_forms = true`

As a sidenote: we're currently hard at work refactoring Spina to use [Hotwire](https://hotwire.dev) which will fix this issue.

wakproductions:
As someone trying to set up a Spina site from a blank Rails 6.1 project, this was unexpected and confusing. I recommend that your documentation setup guide on the [Install Spina CMS page](https://www.spinacms.com/guide/getting-started/installing-spina) note these prerequisites - that you need to:
1) Install ActiveStorage by running the `rails activestorage:install` command
2) Include `gem 'image_processing', '~> 1.2'` in the Gemfile

An early warning on that documentation page would probably help other users avoid discovering this by running into the error as I did.

Also, having tried building a site from a fresh Rails 6.1 app, my experience was that by default Rails _did not_ force `form_with` to be remote. I had to go into the code and modify it to get my image uploader to work. This is why I suggest that you explicitly make that declaration on the `form_with` call itself.

wakproductions:
I removed the `image_processing` reference from this PR, but suggesting explicit remote declarations on the `form_with` calls.

Bramjetten (approved):

PR #656, opened 2021-02-06
Better path generation - Fixes #655

### LINKED ISSUE #655: Wrong materialized path on (resource) pages with same title
I discovered that Spina generates the same materialized path for resource pages with the same title if more than 2 pages share the same title.

Example:

Page 1: 
**title**: Example
**materialized path**: `/slug/example`

Page 2:
**title**: Example
**materialized path**: `/slug/example-1`

Page 3:
**title**: Example
**materialized path**: `/slug/example-1`

In fact, the counter is never incremented.

Bramjetten:
We previously only checked for duplicate localized_material_paths and added a `-[count]` at the end. But because of the way it was programmed that only works once (adding -1). Fixed it using a while loop. Added tests as well.

### DIFF
--- a/app/models/spina/page.rb
+++ b/app/models/spina/page.rb
@@ -126,7 +126,7 @@ module Spina
       end
       
       def duplicate_materialized_path?
-        self.class.where.not(id: id).i18n.where(materialized_path: materialized_path).count > 0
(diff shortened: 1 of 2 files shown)


simmerz (approved):
lgtm 👍

PR #667, opened 2021-02-26
Add resolve routes to main app and replace url helpers - Fixes #666

### LINKED ISSUE #666: Rails 6.1 ActiveStorage "proxy" controller not supported by the Content Presenter
In Rails 6.1, we can set a configuration option like so in application or environment config files:

```
config.active_storage.resolve_model_to_route = :rails_storage_proxy
```

Spina's Content Presenter uses the rails_blob_representation_url which always routes to the redirect controller.

https://github.com/SpinaCMS/Spina/blob/b1f6fbef280170c6e7043954f85d89798c8c2dcd/app/presenters/spina/content_presenter.rb#L45

Leveraging Rail's image_tag helper would probably be easiest since it respects the proxy configuration directive, but Rails seems to have a problem with that with mounted Engines (https://github.com/rails/rails/issues/31325).

I'm not sure if Active Storage provides a more generic url_helper for blobs that would respect the configuration. A simple "rails routes | grep active_storage" doesn't seem to offer any. There is some discussion of a slightly related issue here: https://github.com/rails/rails/issues/41172

Bramjetten:
Spina CMS used a couple of workarounds to deal with url generation for ActiveStorage. Rails 6 fixed an underlying issue, so now we're able to leverage `main_app.url_for` to generate all URL's. 

By adding three `resolve` calls to the main app's routes we can now directly pass `Spina::Parts::Image` and `Spina::Parts::Attachment` objects to the `main_app.url_for` helper. This also enables Spina to use Rails 6.1's new :rails_storage_proxy config (undocumented right now).

ActiveStorage generates redirecting URLs by default. These are hard to cache when using a CDN, so you can add the following line to your app config:
`config.active_storage.resolve_model_to_route = :rails_storage_proxy`

This will make sure the generated URLs do not redirect, but instead download the file directly from your service. Making it cacheable.

This PR also increases the Rails requirement to >= 6 and adds the image_processing gem to the gemspec.


corporealfunk:
I just tested my project which specifies ```config.active_storage.resolve_model_to_route = :rails_storage_proxy``` against this branch and it works, both frontend and the admin panel.

Bramjetten:
Added a check for `resolve_model_to_route` in the routes file so this PR doesn't break 6.0 support.

simmerz (approved):

PR #682, opened 2021-04-07
ActiveStorage variant definitions use ImageProcessing gem macros

corporealfunk:
### Context

ActiveStorage now uses the ImageProcessing gem which allows the use of libvips or imagemagick for variant processing via the config directive config.active_storage.variant_processor

Instead of using imagemagick-specific resize and crop flags, use the macros provided by ImageProcessing so that users of Spina can set :vips as their variant processor.

### Guidance to review

Review macros being used.

### DIFF
--- a/app/helpers/spina/images_helper.rb
+++ b/app/helpers/spina/images_helper.rb
@@ -8,7 +8,7 @@ module Spina
 
     def embedded_image_url(image)
       return "" if image.nil?
-      main_app.url_for(image.variant(resize_to_fit: Spina.config.embedded_image_size))
+      main_app.url_for(image.variant(resize_to_limit: Spina.config.embedded_image_size))
     end
 
   end
--- a/app/helpers/spina/images_helper.rb
+++ b/app/helpers/spina/images_helper.rb
@@ -8,7 +8,15 @@ module Spina
 
     def embedded_image_url(image)
       return "" if image.nil?
-      main_app.url_for(image.variant(resize_to_limit: Spina.config.embedded_image_size))
+
+      # support both ImageProcessing gem macro :resize_to_limit and ImageMagick-specific :resize
+      resize_key = if !Spina.config.embedded_image_size.is_a?(Array)
+                     :resize
+                   else
+                     :resize_to_limit
+                   end
+      byebug
+      main_app.url_for(image.variant(Hash[resize_key, Spina.config.embedded_image_size]))
     end
 
   end
--- a/app/helpers/spina/images_helper.rb
+++ b/app/helpers/spina/images_helper.rb
@@ -15,7 +15,6 @@ module Spina
                    else
                      :resize_to_limit
                    end
-      byebug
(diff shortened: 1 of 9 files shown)


corporealfunk:
Note: this will break anybody's install who has specified Spina.config.embedded_image_size in their project with an imagemagick resize string.

Bramjetten:
Great job!

Any idea how we can handle the change in the initializer? It’s unfortunate that the default included the old directive.

corporealfunk:
I don't think there's any ideal way. A few options, none great:

**Make it backwards compatible:**

1. Spina.config.embedded_image_size is only referenced once in the codebase. The new macros require passing arrays to hold the arguments. If the user is still setting a string, perhaps a log warning could be generated and the code generates an imagemagick style variant definition. If it's an array, use the ImageProcessing macro.
2. How many variations are available to ImageMagick's resize flag? I know there are things like: "2000x2000>", or "2000x2000^". If there aren't too many, maybe write a quick parser that converts the resize string into an array for the new macro. Could be tested extensively. Maybe throw a Log warning that using an array is the new normal. A quick scan of their docs (https://legacy.imagemagick.org/script/command-line-processing.php#geometry) doesn't seem like this is a viable path.

**Break It**

If Spina detects that embedded_image_size is set to anything but an array throw an exception with details about what is happening and why, and how to fix it. Unfortunately, Spina 2.x is already out and this would be a breaking change, so not exactly semver friendly.

corporealfunk:
Ack, I muddled up this PR with another PR I just made regarding media library data-remote: true.

Bramjetten:
I just noticed that this configuration isn't actually present in the generated initializer. Luckily that means most people won't even know about this config option. In that case I'd prefer adding a log warning and use the string variant definition. Would very much like to prevent breaking changes before releasing the hotwire branch.

corporealfunk:
Latest commit reopens the Spina module to redefine the config method that returns the configuration object. Perhaps there's a more elegant way? I wanted to throw the deprecation warning at assignment rather than access.

The helper method than choose which resize key needs to be used based on the object type of the configuration setting.

app/helpers/spina/images_helper.rb
```diff
@@ -8,7 +8,15 @@ def thumbnail_url(image)
...
       return "" if image.nil?
-      main_app.url_for(image.variant(resize_to_limit: Spina.config.embedded_image_size))
+
+      # support both ImageProcessing gem macro :resize_to_limit and ImageMagick-specific :resize
+      resize_key = if !Spina.config.embedded_image_size.is_a?(Array)
+                     :resize
+                   else
+                     :resize_to_limit
+                   end
+      byebug
```
corporealfunk: oops! removed in next commit, I could probably rebase this.

Bramjetten:
Thanks for your updates! Looking good to me. Perhaps you could use `ActiveSupport::Deprecation` to add the deprecation warning. Old config support can be removed in Spina v2.2.

Bramjetten (approved):

Bramjetten:
Thanks @corporealfunk !

PR #688, opened 2021-04-23
Hotwire

Bramjetten:
All JavaScript, CSS, and views have been refactored.

Originally, Spina was built using Haml, custom Sass, and lots of jQuery. That's all gone now. Spina is now built using these fancy new technologies:
- Turbo 7
- Stimulus without Webpacker
- TailwindCSS
- ViewComponent

These changes mean that all Spina plugins will need to be updated to work with the new views. Spina includes a couple of useful view components that you can use for this. 

## Screenshots
This new version of Spina looks a bit different.

CleanShot 2021-04-23 at 11 43 52@2x
CleanShot 2021-04-23 at 11 44 43@2x
CleanShot 2021-04-23 at 11 44 55@2x


corporealfunk:
Found a bug in the repeater part UI. Not sure if the PR comments are the best place for that? When using a Repeater Part with an Image, when I add more than one repeater sub-part and choose images for all of them, only the first Repeater sub-part is updated in the UI:

repeater_bug

Bramjetten:
Good find! Will add a fix here.

Bramjetten:
@corporealfunk This was a bit tricky. When adding new fields in a repeater, we have to generate new ID's so we can target the right input fields. Nested fields (image/text forms) mess this up. I've created a new Stimulus unique-id-controller. This controller makes sure that all occurrences of `f.object.object_id` inside the image/text `_form` partials are always replaced by a new UUID.

```javascript
import { Controller } from "stimulus"

export default class extends Controller {
  
  connect() {
    this.replaceHTML()
  }
  
  replaceHTML() {
    let html = this.element.innerHTML
    
    let uuid = this.generateUUID()
    let regex = new RegExp(this.id, 'g')
    let replaced_html = html.replace(regex, uuid)
    
    this.element.innerHTML = replaced_html
    this.element.id = uuid
  }
  
  generateUUID() {
    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
      var r = Math.random() * 16 | 0, v = c == 'x' ? r : (r & 0x3 | 0x8)
      return v.toString(16)
    })
  }
  
  get id() {
    return this.element.dataset.uniqueId
  }
  
}
```

Inside the partial:
```html
<div class="mt-6 relative" data-controller="media-picker unique-id" data-unique-id="<%= dom_id(f.object, f.object.object_id) %>">
```

simmerz (approved):

PR #695, opened 2021-05-10
Adding  Spina::Parts:Base#label

kbuckler:
### Context
This PR adds a `label` method to `Spina::Parts::Base`. This allows custom parts to define their own administrative labels instead of calling `content#to_s` which may not be applicable in some cases. 

For example:

``` ruby
module MyApp
  module CustomParts
    class Product < Spina::Parts::Base
      attr_json :product_id, :integer

      def content
        Product.find(product_id)
      end

      def label
        content&.name
      end
    end
  end
end
```

I've defined `Spina::Parts::Base#content` for testing purposes but additionally to indicate that this method is expected to be present in subclasses.

### DIFF
--- a/app/models/spina/parts/base.rb
+++ b/app/models/spina/parts/base.rb
@@ -7,6 +7,12 @@ module Spina
 
       attr_json :title, :string
       attr_json :name, :string
+
+      def label
+        content&.to_s
+      end
+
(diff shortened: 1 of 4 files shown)


Bramjetten (approved):
Thanks for your contribution!

PR #707, opened 2021-05-26
Reintroduce configurable tabs for admin/pages

kbuckler:
### Context

Previously, tabs were configurable by 3rd-party plugins by modifying the `@tabs` variable in `Spina::Admin::PagesController`. This PR reintroduces that functionality.

This is primarily a refactoring to reintroduce the `@tabs` variable and `set_tabs` controller filter.

### Guidance to review

1. Please note that `seo` and `search_engines` had been used interchangeably. This PR consolidates on `search_engines`.
2. The search engines tab button had some additional markup that has been removed in this PR. This was removed in favor of being consistent with the other two buttons. I am not sure if that markup is necessary - feedback appreciated.


Bramjetten (commented):
Thanks for your PR! 🙏

What is the @locale variable used for in the TabButtonComponent?

kbuckler:
Ah, the locale argument is not needed! Good catch. Removed.

app/views/spina/admin/pages/_form.html.erb
```diff
@@ -4,67 +4,36 @@
...
-            <label class="font-medium text-gray-700">
-              <%= Spina::Page.human_attribute_name(:title) %>
-            </label>
-          
-            <%= render Spina::Forms::TextFieldComponent.new(f, :title, size: "lg") %>
-            
-            <turbo-frame id="page_content" src="<%= spina.edit_content_admin_page_path(f.object, locale: @locale) %>">
-            </turbo-frame>
+        <% @tabs.each do |tab_name| %>
+          <div data-tabs-target="pane" id="<%= tab_name %>" <%= 'hidden' unless tab_name == @tabs.first %>>
```
Bramjetten, line 32: Very minor nitpick, but you could also write this as:

```erb
<% @tabs.each.with_index do |tab_name, index| %>
  <div data-tabs-target="pane" id="<%= tab_name %>" <%= "hidden" unless index.zero? %>>
```

kbuckler:
Any thoughts on testing this and other view components in general?

app/views/spina/admin/pages/_form.html.erb
kbuckler: Sounds good, changed!

app/components/spina/pages/tab_button_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<button type="button" class="block px-3 leading-relaxed py-1 hover:text-gray-800 rounded-md text-gray-400 font-medium text-sm flex items-center whitespace-nowrap" data-action="tabs#show" data-tabs-target="button" data-pane-id="<%= @tab_name %>">
+  <%= helpers.heroicon(icon_name, class: 'w-4 h-4 mr-1 -ml-1 opacity-75') %>
```
Bramjetten, line 2: Instead of always including this specific heroicon helper, I think it's better to explicitly set this button content where it's called. That way we don't have to hard-code icons in the component class. It also enables buttons without icons. Kinda like the `Spina::UserInterface::DropdownComponent`

app/views/spina/admin/pages/_form.html.erb
```diff
@@ -4,67 +4,36 @@
...
-            <%=t 'spina.pages.page_content' %>
-          </button>
-          <button type="button" class="block px-3 leading-relaxed py-1 rounded-md hover:text-gray-800 text-gray-400 font-medium text-sm flex items-center whitespace-nowrap" data-action="tabs#show" data-tabs-target="button" data-pane-id="seo">
-            <%= heroicon('search', style: :solid, class: 'w-4 h-4 mr-1 -ml-1 opacity-75') %>
-            <span class="hidden md:inline">
-              <%=t 'spina.pages.search_engines' %>
-            </span>
-            <span class="md:hidden">
-              SEO
-            </span>
```
Bramjetten, line 31: This markup was used to show a shorthand in the mobile view. We can let the `TabButtonComponent` render just the `<button>` element, and keep this in when we need it.

app/components/spina/pages/tab_button_component.html.erb
kbuckler: Yeah, I agree. However, I couldn't find a better place to keep the mapping between the tab name and the icon filename.

Bramjetten:
I think Spina could benefit from proper system tests (instead of just model/controller tests right now). That's out of the scope of this PR though.

app/components/spina/pages/tab_button_component.html.erb
kbuckler: Additionally, I'm also looking for an easily overridable entry point to set the tab icon from an external plugin.

kbuckler:
@Bramjetten Perhaps instead of using the TabButtonComponent here, I should just render each in its own partial similar to how the panes are rendered? That will allow for customization without much overhead.

Bramjetten:
I’m sorry, I misunderstood how this works. Your first proposal is the way we should go I think. It’s a shame we need to hard code those icons, but that’s fine for now imo

kbuckler:
No problem - I'm actually close to getting working as you've suggested and I *think* it's actually a better approach, even if it's a little verbose. 

For example, in the Solidus Spina plugin adding a new tab requires two partials:

Screen Shot 2021-05-26 at 2 28 39 PM

kbuckler:
Ok, here's another approach. New tabs can be added tabs by modifying`@tabs` in the controller and adding the two corresponding partials, `_button_#{tab_name}` and `_form_#{tab_name}`. 

Appreciate your feedback @Bramjetten!

app/views/spina/admin/pages/_form.html.erb
kbuckler: Readded this with my last commit.

Bramjetten:
One more suggestion: instead of setting separate icon/mobile/label variable, just have one content slot so we can create buttons like this:

```ruby
<%= render Spina::Pages::TabButtonComponent.new(tab_name: 'advanced') do |button| %>
  <% button.content do %>
    <%= heroicon('cog', style: :solid, class: 'w-4 h-4 mr-1 -ml-1 opacity-75') %>
    <%= t("spina.pages.advanced") %>
  <% end %>
<% end %>
```

This way plugins can pretty much add anything they want!

Bramjetten:
One more suggestion: instead of setting separate icon/mobile/label variable, just have one content slot so we can create buttons like this:

```ruby
<%= render Spina::Pages::TabButtonComponent.new(tab_name: 'advanced') do |button| %>
  <% button.content do %>
    <%= heroicon('cog', style: :solid, class: 'w-4 h-4 mr-1 -ml-1 opacity-75') %>
    <%= t("spina.pages.advanced") %>
  <% end %>
<% end %>
```

This way plugins can pretty much add anything they want!

kbuckler:
Yeah, that makes more sense now that each button has its own partial.

Bramjetten (approved):
Thanks again for working on this!

PR #708, opened 2021-05-28
Update ViewComponent and change content slot to label

Bramjetten:
Updated ViewComponent to 2.32. Removed `include ViewComponent::SlotableV2` because it's not required anymore.

The new `TabButtonComponent` that was introduced in #707 used a slot name `content` which is not supported anymore, so I changed that to `label`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -15,7 +15,7 @@ PATH
       sass-rails
       stimulus-rails
       turbo-rails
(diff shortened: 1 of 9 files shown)


Bramjetten:
I simplified the `TabButtonComponent` by removing the single `label` slot and instead just rendering `content`.

simmerz (approved):

PR #716, opened 2021-06-04
Only create rewrite_rule on update

Bramjetten:
Spina currently creates rewrite rules even when pages are first created. That's not ideal because it creates a rewrite rule with `nil` as the `old_path`. 

This PR moves the AR callback to `after_update`.

### DIFF
--- a/app/models/spina/page.rb
+++ b/app/models/spina/page.rb
@@ -35,7 +35,7 @@ module Spina
     after_save :touch_navigations
 
     # Create a 301 redirect if materialized_path changed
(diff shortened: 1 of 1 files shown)


kbuckler:
👍  A minor suggestion: perhaps explicitly defining what to do with `nil` paths in `RewriteRule` or `Spina::Page#rewrite_rule` more would assist with clarity/readability.

Bramjetten:
I think both old_path and new_path should have a NOT NULL constraint in the db. Nil values don’t make any sense for rewrite rules.

PR #719, opened 2021-06-10
Hide user management buttons for normal users

kahrpatrick:
### Context
Buttons for adding and editing users are not hidden for normal users. Since only admins can manage users, this leads to an error.

Check if the current user is admin before displaying user management buttons.


Bramjetten (approved):

Bramjetten:
Note to self: implement proper policies using pundit.

Bramjetten (approved):

PR #720, opened 2021-06-10
Save all pages after updating resource

### LINKED ISSUE #714: changing resource slug does not update existing pages
When changing the slug of a resource via the settings UI, existing pages of that resource are not updated.

kahrpatrick:
### Context
Fixes #714 

Update all pages of a resource when it is saved to make sure the materialized path is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Refactored all views with ViewComponent
 * Refactored all javascript with Hotwire
 * Added importmap helper `spina_importmap_from`
+* Added background job to save resource pages in bulk
 
 ## 2.0
 
--- a/app/jobs/spina/application_job.rb
+++ b/app/jobs/spina/application_job.rb
@@ -1,0 +1,9 @@
+module Spina
+  class ApplicationJob < ActiveJob::Base
+    # Automatically retry jobs that encountered a deadlock
+    # retry_on ActiveRecord::Deadlocked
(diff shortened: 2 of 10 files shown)


Bramjetten:
I think this is a good idea, but I prefer a different implementation. Can we move this to a bg job? I have projects with several resources containing 1.000+ pages

kahrpatrick:
Sure, we could set it up as a job. Ideally this would require an external queuing backend to prevent inconsistencies in case of a crash. What do you think about this?

Bramjetten:
I think we could just create an ActiveJob and let the end user/developer decide which queuing backend they want to use. Kinda like how ActiveStorage creates activejobs for analyzing and purging files.

kahrpatrick:
Implemented as a job now.

Bramjetten:
Thanks! 

I've added configurable queues to Spina. You can configure how to queue these jobs:

```ruby
Spina.configure do |config|
  config.queues.page_updates = :your_queue
end
```

It defaults to your queue adapter's default.

Bramjetten:
I've also changed the callback to `after_commit`. Sidekiq is sometimes too fast otherwise!

kahrpatrick:
Thanks, makes sense.

Bramjetten (approved):

Bramjetten:
@kahrpatrick Thanks for your work!

PR #727, opened 2021-06-24
Clear turbo frame src - fixes #726

### LINKED ISSUE #726: Media Picker modal doesn't reopen if closed without choosing an image
When I click "Choose Image" from the Spina::Image part UI from a page, the media picker modal opens. If I do not choose any images, and instead press "ESC" or click outside of the modal area, the modal closes.

However, if click again on "Choose Image", the modal does not open again.

(If I click on "Choose Image" on a different Spina::Image the modal will open for that target. And then the first target is valid again.)

media_picker_bug

Bramjetten:
After a form submission in a turbo frame, you cannot return a 200 response. You have to use a redirect. 

Closing a modal now also resets the `src` attribute.

### DIFF
--- a/app/assets/javascripts/spina/controllers/modal_controller.js
+++ b/app/assets/javascripts/spina/controllers/modal_controller.js
@@ -8,10 +8,15 @@ export default class extends Controller {
 
   close() {
     this.element.remove()
+    this.modalTurboFrame.src = null
   }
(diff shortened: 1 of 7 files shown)


corporealfunk:
Checked out the branch on my project and it looks good to me.

PR #741, opened 2021-07-04
Fixes pages custom path if Spine Engine route is different than default

tellodaniel:
### Context

I want to change Spina root url since I'm using it in a existing project. I changed `routes.rb` configuration with:

`mount Spina::Engine => '/blog'`

After this change Pages urls are broken since they are missing root path (`/blog`)

Add a Page method (`url`) to include Spina::Engine root url. This change affects:

1. `Spina::MenuPresenter`
2. Admin pages views

### Guidance to review

Change `routes.rb` configuration and check if all Pages urls are working correctly

(This is my first day using Spina. Sorry if this solution is not the best)

### DIFF
--- a/app/models/spina/page.rb
+++ b/app/models/spina/page.rb
@@ -55,6 +55,10 @@ module Spina
     def slug
       url_title&.parameterize
     end
+
+    def url
+      [engine_path, materialized_path].join
+    end
     
     def homepage?
       name == 'homepage'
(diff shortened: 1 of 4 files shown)


Bramjetten:
We recommend not mounting Spina on a different path. You can disable Spina's frontend routes and add your own if you want. I'd recommend creating a resource `Blog` with a `/blog` slug so that gets prepended to all blog pages. That way your pages will all have a `/blog/your-article` URL without having to change Spina's internals.

tellodaniel:
Thanks @Bramjetten for your quick response. Just tested what you recommend and works. The only gotcha is that our homepage (in root path `/`) now must be handled by Spina. We will move it to Spina and see how it works.

Bramjetten:
You can move Spina’s routes to the bottom of your routes file and use your own route for your homepage.

PR #744, opened 2021-07-05
Spina::Authentication

Bramjetten:
Extracted out authentication so it's easier to swap it out for something like Devise


simmerz (changes requested):
I like what's here, but I'd like to see a test against the default auth, and separately a test using the custom `session[:user_id]` module so we don't end up with issues down the line.

Bramjetten:
So... I've been working on tests for this, but it looks like it's quite hard to test this properly. Changing `config.authentication` and reloading `Spina::ApplicationController` apparently doesn't cut it. Not sure how to tackle this.

Bramjetten:
Took some effort, but included tests for different auth strategies!

simmerz (approved):

PR #753, opened 2021-07-22
add links to spina attachments index

derschiw:
### Context
Hello!
We very much like spina and use it a lot in our projects. For a current project people wanted to be able to add links to uploaded pdf files (e.g. in the Trix editor). Currently the attachments can only be embedded as a part in a theme and there is no way to get a link to the blob. Also this is a easy way to via at least pdf files in the browser.

Add a link to the attachments.

### Guidance to review
Its just a single line with a link.

### DIFF
--- a/app/views/spina/admin/attachments/_attachment.html.erb
+++ b/app/views/spina/admin/attachments/_attachment.html.erb
@@ -2,7 +2,7 @@
   <div class="flex items-center h-12 border-b border-gray-200 px-8 hover:bg-white group">
     
     <turbo-frame id="<%= dom_id(attachment) %>_filename" class="font-medium text-gray-800 h-full flex-1 w-56 text-sm flex items-center relative">
-      <span class="truncate"><%= attachment.file.filename %></span>
+      <span class="truncate"><%= link_to attachment.file.filename, file_url(attachment.file) %></span>
     
       <div class="absolute h-full opacity-0 group-hover:opacity-100 flex items-center right-0 pr-3 top-0">
         <%= link_to spina.edit_admin_attachment_path(attachment), class: "btn btn-default h-7 px-2 mr-2 text-xs" do %>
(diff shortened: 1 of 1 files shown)


app/views/spina/admin/attachments/_attachment.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="flex items-center h-12 border-b border-gray-200 px-8 hover:bg-white group">
     
     <turbo-frame id="<%= dom_id(attachment) %>_filename" class="font-medium text-gray-800 h-full flex-1 w-56 text-sm flex items-center relative">
-      <span class="truncate"><%= attachment.file.filename %></span>
+      <span class="truncate"><%= link_to attachment.file.filename, file_url(attachment.file), target: :_blank %></span>
```
Bramjetten, line 5: Can we add some styles so these links are easier to click?

```erb
<%= link_to file_url(attachment.file), target: :blank, class: 'block h-full w-full flex items-center truncate hover:underline' do %>
  <%= attachment.file.filename %>
<% end %>
```

Bramjetten:
The helper method `file_url` still uses the `rails_service_blob_path`, but I think it's better to just use `main_app.url_for`. Could you change that?

Bramjetten:
Thanks for your PR!

Bramjetten:
I was playing around with this some more, and wouldn't it be nicer to add a separate download button? I think it makes it more clear to end users:

CleanShot 2021-07-22 at 15 59 27@2x

derschiw:
Sure this seems nice!

derschiw:
> The helper method `file_url` still uses the `rails_service_blob_path`, but I think it's better to just use `main_app.url_for`. Could you change that?

I was not sure, if this is a good idea since the don't create permanent links AFAIK?

Bramjetten:
`main_app.url_for` is what Spina's content method uses to create links to attachments. It leaves it up to the main app's ActiveStorage setup to generate the correct URL. `rails_service_blob_path` is not really meant to be used outside of Rails I think

derschiw:
> `main_app.url_for` is what Spina's content method uses to create links to attachments. It leaves it up to the main app's ActiveStorage setup to generate the correct URL. `rails_service_blob_path` is not really meant to be used outside of Rails I think

Ok I could add a helper method for that? Or have you changed it allready?

Bramjetten:
I'll add a commit to this PR if you can give me write access

Bramjetten:
Closing in favor of #754

PR #758, opened 2021-08-01
API

Bramjetten:


app/serializers/spina/api/image_serializer.rb
```diff
@@ -0,0 +1,11 @@
+module Spina::Api
+  class ImageSerializer < BaseSerializer    
+    set_type :image
+    
+    attributes :id
+    
+    attribute :url do |image, params|
+      params[:view_context].main_app.url_for(image.variant(resize_to_fill: [2000, 2000]))
```
simmerz, line 8: I'd probably provide the original url, a resize_to_fit and a square thumb. 2000x2000, resized to fill will easily result in the wrong aspect ratio for some sites.

Bramjetten: I've added an `original_url`, `embedded_image_size_url` (default: 2000x2000) and `thumbnail_url` (default: 400x400). Those last two are configurable. Will add on-the-fly transformations in the future.

simmerz (approved):

PR #879, opened 2021-10-29
Use Spina.mounted_at to generate paths when not mounted at root

Bramjetten:
Since the beginning, Spina has always assumed it was mounted at `/`. This PR adds support for other paths.

Spina uses `Spina::Engine.routes.find_script_name({})` to find the path it's mounted at. Rails doesn't include a nice method for getting an engine's mount path, so I'm not 100% sure about this. An alternative would be to add a configurable setting in `spina.rb` to set the mount path, but then you'll have to make sure that config is always in sync with your routes file.

A warning though: once set, all paths will be generated using the mount location. That means that if you change it later, you'll have to regenerate all page paths.

### DIFF
--- a/app/controllers/concerns/spina/frontend.rb
+++ b/app/controllers/concerns/spina/frontend.rb
@@ -53,7 +53,7 @@ module Spina
       end
 
       def spina_request_path
-        segments = ['/', params[:locale], params[:id]].compact
+        segments = [Spina.mounted_at, params[:locale], params[:id]].compact
         File.join(*segments)
       end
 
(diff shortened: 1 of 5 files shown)


simmerz:
Is it worth raising an issue on https://github.com/rails/rails to add a method to get the current engine's mount path?

Bramjetten:
I can't find the original issue, but it was proposed to Rails and turned down because of all kinds of nasty bugs.

simmerz:
Ah ok. Should we add some documentation as to how to regenerate paths (and that you have to) following a change in mount path?

Bramjetten:
Yes, I’ll add it to this PR 👍

Bramjetten:
@simmerz Added documentation

PR #881, opened 2021-10-31
[enhancement] silent installation

bdavidxyz:
### Context

Enhancement proposal : accept an optional silent install.

The user can optionally ask for a silent install when first installing Spina on a new app.

### Guidance to review

Try to install Spina from this branch on a fresh new Rails 6 app, with the "--silent" option when doing "bin/rails spina:install". Like this :

```
bin/rails spina:install --silent
```

If Spina is correctly installed with default values, then the PR is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,14 +20,7 @@ Run the installer to start the setup process:
 
     rails g spina:install [--silent]
 
-The installer will help you create your first user. 
-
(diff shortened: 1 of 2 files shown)


Bramjetten:
Thank you for your PR 🙏 Much appreciated!

Bramjetten (approved):

PR #968, opened 2022-02-14
Add active_list_item_css to Menu Presenter

sarabpj:
### Context
 @Bramjetten had a conversation with @mweitzel in discord about adding styles to the list item in Menu Presenter that matches current page. 

Add `active_list_item_css` that applies config styles to list item that matches current page

### DIFF
--- a/app/presenters/spina/menu_presenter.rb
+++ b/app/presenters/spina/menu_presenter.rb
@@ -88,8 +88,8 @@ module Spina
       end
 
       def parent_of_current?(item)
-        return false unless Spina::Current.page.ancestry.present?
-        Spina::Current.page.ancestry.split('/').map(&:to_i).include? item.id
+        return false if item.homepage?
+        Spina::Current.page.materialized_path.starts_with? item.materialized_path
       end
   end
 end
--- a/app/presenters/spina/menu_presenter.rb
+++ b/app/presenters/spina/menu_presenter.rb
@@ -8,9 +8,10 @@ module Spina
 
     # Configuration
     config_accessor :menu_tag, :menu_css,
-                    :list_tag, :list_css, 
-                    :list_item_tag, :list_item_css, 
+                    :list_tag, :list_css,
+                    :list_item_tag, :list_item_css,
                     :link_tag_css,
+                    :active_list_item_css,
                     :include_drafts,
                     :depth # root nodes are at depth 0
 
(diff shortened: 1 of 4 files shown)


Bramjetten:
Thank you for this PR!

I see you've only added an active class to the current page. It might be useful to style parent pages differently as well.
How about `active_list_item_css` for all ancestors, and `current_list_item_css` for the current page?

Comparing all ancestors to the current page would create N+1 queries though, so then it might be better to compare to the current path to `materialized_path` instead.

sarabpj:
@Bramjetten  just to make sure I understand the request correctly `active_list_item_css ` should apply a class on the body of the current page?

Bramjetten:
@sarabpj Imagine a navigation tree like this:

- Home
- About
- - Company
- - History

When visiting /about/history, `active_list_item_css` should apply a class to the About page, and `current_list_item_css` should apply a class to the History page.

sarabpj:
Hey @Bramjetten I went ahead and made the changes you requested please let me know if there is anything else.

Bramjetten:
Thanks for your updates! 🙏

You're currently using `ancestry` to check parent pages. While this definitely works, it also creates a lot of SQL-queries. Especially in a large menu. I'd prefer to compare the materialized path instead. That way we don't have to fetch all parents recursively.

Maybe something like this would suffice:

```ruby
def parent_of_current?(item)
  return false if item.homepage?
  Spina::Current.page.materialized_path.start_with? item.materialized_path
end
```

sarabpj:
Yes, that works 😁

Bramjetten:
Again, thank you for working on Spina. Much appreciated 🙏 

I refactored the code a little bit and added a small test to make sure the `MenuPresenter` behaves like it should.
<sup>Our tests are *very* brittle on Github Actions and I'm not sure why... Tests are passing locally... 🤔</sup> 

Merging!

Bramjetten (approved):

kfrz:
I believe this change breaks custom navigation rendering; as described in the issue I filed today - https://github.com/SpinaCMS/Spina/issues/1164. 

I am continuing investigation and will comment any further findings.

kfrz:
Submitted a small PR that, at least for me, fixes this issue. I did *not* write any tests, because I'm feeling lazy this afternoon. Sorry. https://github.com/SpinaCMS/Spina/pull/1165

PR #979, opened 2022-02-25
Fixing silent install generator flag and adding rake task

tobiasbhn:
### Context
In order to Dockerize my Rails Application using Spina CMS, i need to automatically setup my database using `rails generate spina:install --first-deploy --silent`, but as a rake task (no rails context available in docker image). The `--first-deploy` flag already has been setup as a rake task, the `--silent` flag not yet. 

This PR
* adds the rake task `rake spina:silent_first_deploy`, to initialize the database without any console prompts.
* fixes the `--silent` flag itself, which outputs prompts even if silent flag is set (if database already set up)

### DIFF
--- a/lib/generators/spina/install_generator.rb
+++ b/lib/generators/spina/install_generator.rb
@@ -32,7 +32,7 @@ module Spina
     end
 
     def create_account
-      return if ::Spina::Account.exists? && !no?('An account already exists. Skip? [Yn]')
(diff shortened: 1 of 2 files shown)


Bramjetten (approved):

tobiasbhn:
This PR actually has brought a simple bug. See new PR #980 for fix.

PR #989, opened 2022-03-08
Addresses common namespace conflict for current_account method

moxvallix:
We are implementing Spina into our existing Rails project.
We've come across a namespace conflict with the current_account method.

We've found its increasingly common to have current_user and current_account defined within the authentication model and Spina in using the same method name is causing us a conflict as we attempt to integrate Spina into our broader applications authentication.

To assist users to avoid this obvious potential conflict we'd like to offer that in spina, current_account be renamed to current_spina_account, freeing up the Account model to cause no conflicts in the clients main app.

We've created a pull request that makes this single change across the code base. All tests pass.
And the change is tested to be working in our app that integrates Spina.

### DIFF
--- a/app/controllers/concerns/spina/current_spina_account.rb
+++ b/app/controllers/concerns/spina/current_spina_account.rb
@@ -1,15 +1,15 @@
 module Spina
-  module CurrentAccount
+  module CurrentSpinaAccount
     extend ActiveSupport::Concern
     
     included do
-      before_action :current_account
-      helper_method :current_account
+      before_action :current_spina_account
+      helper_method :current_spina_account
     end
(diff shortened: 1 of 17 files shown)


Bramjetten:
Thank you for your PR!

I understand and agree that this method should be named `current_spina_account`. However, this change would break existing projects using `current_account`. Perhaps we could re-add the `current_account` method as an alias to `current_spina_account`, and add a deprecation warning.

moxvallix:
Great idea! Looking into deprecation warnings now, will come back when we have implemented them.

moxvallix:
moved to #991

PR #991, opened 2022-03-09
Change current_account to current_spina_account, and add deprecation warning.

moxvallix:
I moved this to a new branch in my repo, and had to create a second PR.

current_account is now an alias of current_spina_account, and a deprecation warning has been added to it.

See #989 for old PR.

### DIFF
--- a/app/controllers/concerns/spina/current_spina_account.rb
+++ b/app/controllers/concerns/spina/current_spina_account.rb
@@ -5,17 +5,19 @@ module Spina
     included do
       before_action :current_spina_account
       helper_method :current_spina_account
-      helper_method :current_account unless Spina.disable_deprecated_methods
+      helper_method :current_account
     end
   
     private
 
-      def current_account
-        ActiveSupport::Deprecation.warn(
-          "#current_account is deprecated, due to a common authentication namespace conflict. \n" \
-          "Please use #current_spina_account instead."
-        )
-        current_spina_account
+      unless Spina.disable_deprecated_methods
+        def current_account
+          ActiveSupport::Deprecation.warn(
+            "#current_account is deprecated, due to a common authentication namespace conflict. \n" \
+            "Please use #current_spina_account instead."
+          )
+          current_spina_account
+        end
       end
 
       def current_spina_account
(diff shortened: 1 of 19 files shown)


moxvallix:
I have added an extra config option to the initializer, namely "disable_deprecated_methods".
If the option is true, "current_account" is disabled.

The option defaults to false, and an entry to the initializer template has been made, with the option to set it to true commented out, as well as a hint to uncomment it if they are experiencing a namespace conflict. 

I have tested it in a blank rails app and it is all working, and all tests are passing on my machine.

Bramjetten:
Thank you for your PR. Looks good, but needs a little bit of tweaking.

I think we should come up with a more descriptive configuration option, `disable_deprecated_methods` is too general imo. How about something like `disable_legacy_current_account_method`?

app/controllers/concerns/spina/current_spina_account.rb
```diff
@@ -0,0 +1,28 @@
...
+    
+    included do
+      before_action :current_spina_account
+      helper_method :current_spina_account
+      helper_method :current_account
+    end
+  
+    private
+
+      unless Spina.disable_deprecated_methods
```
Bramjetten, line 13: Shouldn't this be `Spina.config.disable_deprecated_methods`?

app/controllers/concerns/spina/current_spina_account.rb
```diff
@@ -0,0 +1,28 @@
+module Spina
+  module CurrentSpinaAccount
+    extend ActiveSupport::Concern
+    
+    included do
+      before_action :current_spina_account
+      helper_method :current_spina_account
+      helper_method :current_account
```
Bramjetten, line 8: Any way we can add the same conditional here?

moxvallix: I tried that originally, putting an unless at the end of `helper_method :current_account` but it seemingly wasn't working that way. I agree just wrapping the whole method in an if statement is ugly, but it was the only way I could see to disable it. Any suggestions on this would be appreciated :)

app/controllers/concerns/spina/current_spina_account.rb
moxvallix: Yeah, I will change that.

Bramjetten:
Still planning on merging this.

Bramjetten:
Merged

PR #1015, opened 2022-03-30
Replace image/attachment files in media library

Bramjetten:
Added the ability to replace an image's or attachment's file.

CleanShot 2022-03-30 at 15 32 52@2x


Bramjetten:
Added a bg job that replaces old signed blob ID's with the new signed blob ID so that the image is replaced on all pages.

tobiasbhn:
Tested, got one little gotcha: regex_replace will only replace first occurrence. But we need all occurrences in case of same image in multiple languages or same image in Spina Part and Trix Rich Text. Should be simple using the global flag. ([StackOverflow](https://stackoverflow.com/questions/19561299/using-postgres-regexp-replace-to-replace-a-list-of-strings-by-always-the-same-st))

tobiasbhn (changes requested):
Tested current changes in PR, looks good on page content. However, Images could also be used in Layout Parts. Therefore, ReplaceJob should also run on Accounts that used that Image

PR #1023, opened 2022-04-08
animated gif's can't have variants

tobiasbhn:
### Context
As an editor, I want to upload animated gifs to the media library, which are still animated when I display them in the frontend. According to [this link](https://stackoverflow.com/a/57250549/14826603), this is not possible with variants.

Similar behavior is used for svgs, which are also not allowed to have variants. The behavior for gifs was added analogously.

### DIFF
--- a/app/models/spina/image.rb
+++ b/app/models/spina/image.rb
@@ -13,6 +13,7 @@ module Spina
     def variant(options)
       return "" unless file.attached?
       return file if file.content_type.include?('svg')
+      return file if file.content_type.include?('gif')
       return file unless file.variable?
 
       file.variant(options)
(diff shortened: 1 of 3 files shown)


Bramjetten:
I don't think it's entirely impossible to resize gifs using a different library (see: https://github.com/rails/rails/issues/37833)

We do that in our own projects, so I'd rather not disable that completely. There's always the option in your own frontend to use the original file.

tobiasbhn:
Ah got that. :+1: 
`<%= content.image_tag :main_image, { loader: { page: nil } }, { class: "rounded drop-shadow" } %>` does indeed work. Great, did not found this information.

PR #1051, opened 2022-06-07
Add customizable tailwind plugins

joshleblanc:
### Context

I originally wrote this before realizing Spina only includes tailwind for its own admin interface, but I think this still holds value for any spina plugins that need to modify the available plugins.

This change show allow modifying the plugins array generated by the tailwind rake class. Plugins are specified under the `tailwind_plugins` config key.

### DIFF
--- a/lib/spina.rb
+++ b/lib/spina.rb
@@ -22,17 +22,18 @@ module Spina
   config_accessor :api_key,
                   :api_path,
                   :authentication,
-                  :backend_path, 
+                  :backend_path,
                   :importmap,
                   :frontend_parent_controller,
                   :disable_frontend_routes,
                   :disable_decorator_load,
-                  :locales, 
+                  :locales,
                   :embedded_image_size,
                   :mailer_defaults,
                   :thumbnail_image_size,
                   :party_pooper,
                   :tailwind_content,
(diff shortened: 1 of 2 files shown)


Bramjetten:
This is really great, thanks!

Bramjetten (approved):

lib/spina.rb
```diff
@@ -47,20 +48,20 @@ module Spina
   self.mailer_defaults = ActiveSupport::OrderedOptions.new
   self.thumbnail_image_size = [400, 400]
   self.frontend_parent_controller = "ApplicationController"
-  self.locales = []
+  self.locales = [I18n.default_locale]
```
noahsettersten, line 51: @Bramjetten Will this cause Spina to default to `:en` when the initializer is loaded before any changes to the default locale? I saw there was a recent fix for this in #1022 and wasn't sure if this default would conflict with that.

Bramjetten: This slipped through, I believe it simply doesn't do anything.

joshleblanc: Oh my bad, I didn't do that intentionally. I'm not actually sure how that got in there

Issue #1045, opened 2022-05-26
Feature Request: Ability to add resource pages to the navigation


wrystr:
**First of all, I want to say thank you all for bringing this beautiful CMS.**

I can't seem to add resource pages to a navigation, is this expected or are there settings somewhere I've missed?

I would be very grateful if this could happen. Sometimes I need to group pages into resources and display them but only what I want to display in the navigation.

For now I can work around this by using resources to create navigation and limiting it by utilizing the parent page and sub page level 1.

mweitzel:
I've been recycling the MenuPresenter. Locally I have `Spina::ResourceMenuPresenter` with a the `scoped_collection` overriden:
```ruby
module Spina
  class ResourceMenuPresenter < MenuPresenter
    def scoped_collection(collection)
      scoped = collection.active.in_menu.sorted
      include_drafts ? scoped : scoped.live
    end
  end
end
```
and then to use it I'll pass the collection I care about, e.g.
```erb
  <% presenter = Spina::ResourceMenuPresenter.new(current_page.resource.pages) %>
  <% presenter.config.list_css = "delightful-styles" %>
  <% presenter.config.current_list_item_css = "seriously-just-gorgeous" %>
  <% presenter.config.link_tag_css = "ok-maybe-i-have-bias" %>
  <%= presenter.to_html %>
```
So really just dropping the `.regular_pages` scope from the collection might give you what you want.

Bramjetten:
While that's a good solution for creating a navigation menu in your frontend, I think @wrystr means he wants to add resource pages to the Spina::Navigation model from the UI in Spina. On a db-level, there's nothing that prevents us from doing that. The challenge is creating a nice UI for looking up pages and adding them. Resources can contain many (in our case 100s) of pages, making a simple select which we use now unusable. So: we need some sort of fancy PageSelectWithSearchComponent!

PR #1062, opened 2022-06-24
Add actions to create Spina::Resource records

JuanKman94:
## Context

I started using Spina locally and noticed the "collections" model is cryptic. It's required to create new pages but as an user I didn't get any introduction nor instructions on how they work nor why they work as they do.

## Changes proposed in this pull request

* Add navigation links to add new resources/collections (see #1045)
* Add `name` field to `Spina::Resource` form

## Screenshots

image

image

### DIFF
--- a/app/controllers/spina/admin/resources_controller.rb
+++ b/app/controllers/spina/admin/resources_controller.rb
@@ -1,17 +1,40 @@
+# frozen_string_literal: true
+
 module Spina
   module Admin
     class ResourcesController < AdminController
       before_action :set_locale
       before_action :set_resource, only: [:edit, :update]
 
-      def edit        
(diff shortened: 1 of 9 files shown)


Bramjetten:
Hi and welcome! Thank you for working on Spina. 

Collections are not required to make pages. By default, pages belong to the “main” collection (technically meaning *no* collection). 

If you want you can create additional page collections, separate from your main pages. Think: blog posts, news articles or something else which results in a long ordered list of pages. 

These collections are specific to the project and should therefore be defined in the theme initializer by the developer. Users shouldn’t be able to create or delete collections from the UI.

JuanKman94:
Aah, I see. Yeah, I misunderstood #1045 and forgot to RTFM :sweat_smile: Thanks for the response, I'm gonna close this :+1:

PR #1087, opened 2022-08-16
Pagination

Bramjetten:
Adds pagination for pages#index when @pages.count > Spina.config.pages_limit_value (default: 25). Manual sorting by dragging and dropping is then replaced by up/down buttons.

https://user-images.githubusercontent.com/423116/184900011-258a504c-804b-4d4e-bf6d-5c701e113da0.mov


simmerz (changes requested):
I think this needs some tests adding. Looks great though!

Don't think it needs much but definitely a system test or two to cover the features here.

Bramjetten:
Added a simple system test and changed pagination to only apply to resource pages. I want pages in the main collection to always be draggable. You shouldn't be making hundreds of pages there anyway.

simmerz (approved):

PR #1096, opened 2022-08-29
Fix /admin/navigations by not running callback on unused action

ugisozols:
Was getting `The show action could not be found for the :set_navigation callback on Spina::Admin::NavigationsController, but it is listed in its :only option` in a fresh application when visiting /admin/navigations.


Bramjetten:
Great seeing you here! Thanks for finding and fixing this 👍

PR #1118, opened 2022-10-11
Deprecate the use of app/decorators

Bramjetten:
Instead of using `app/decorators` and requiring those files from the engine, I propose we follow the [Rails Guides](https://guides.rubyonrails.org/engines.html#overriding-models-and-controllers) and switch to a different setup to override the engine's functionality.

I wrote a small guide about how to best override Spina's classes and methods: [Overriding Spina](https://spinacms.com/guides/plugins/overriding-spina)

### DIFF
--- a/lib/spina/engine.rb
+++ b/lib/spina/engine.rb
@@ -20,9 +20,9 @@ module Spina
     config.autoload_paths += %W( #{config.root}/lib )
 
     config.to_prepare do
-      # Require decorators from main application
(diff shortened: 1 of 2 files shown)


mattpolito:
This makes sense. Having overrides in the form of a decorator is a bit confusing since 'decorator' can be a loaded term.

mattpolito (commented):
Seems clear.

PR #1137, opened 2022-11-09
Back to using the power off icon for logging out

Bramjetten:
I disliked the current logout icon in Heroicons, so I changed it back to the original power-off icon.

CleanShot 2022-11-09 at 10 19 36@2x


mattpolito:
Seems fine to me.

PR #1148, opened 2022-11-30
Page delete redirect

mattpolito:
### Context

When deleting a page the redirect back to `/admin/pages` can break you out of context if the deleted page was within a resource.

Deleting a page will now redirect as previous but if the page had a resource, the redirect will push the user back to the resource silo.


Bramjetten:
Excellent!

Bramjetten (approved):

PR #1151, opened 2022-12-01
Utilize standardrb for code consistency

mattpolito:
### Context

I've noticed things throughout the codebase such as inconsistent spacing. As an open source project I believe it would be easier for all if set of community driven guidelines helped guide new development.

Having much luck with [standardrb](https://github.com/testdouble/standard) on many projects, I would recommend it since it is intended to not have any linter configuration.

I completely understand that adding a linter can be an adjustment (I believe for the better). When I first started using it there were a couple rules that irked me but I learned to embrace them after understanding the reasoning. It's actually made my code better and very quickly readable.

This PR is intended to be a discussion starter and would not be offended if it's not a direction you would like to go. 

I propose that the project uses standardrb as a dev dependency and I've also provided an initial cleanup.

### Guidance to review

There are many altered files here but the majority of them are only whitespace changes. The few times I actually adjusted code was for things like useless assignment of a variable.


Bramjetten:
I have no experience with this, but it looks like something we should implement. Is this something that can be applied automatically to people's PRs?

mattpolito:
@Bramjetten Yes, you can have linting rules applied during the pr process just like you have tests running.

BTW I just saw there is a test failure, I'll take a look at that.

Bramjetten:
Yeah, that's just Github Actions being unpredictable unfortunately.

mattpolito:
Looks like a flicker locally as well. I was just running `rails test` so I didn't see the failure in the system tests before I pushed it up. However, system tests are flickering locally. I'll see if I can remedy that.

Bramjetten (approved):

Issue #1164, opened 2022-12-19
Spina v2.11 `Spina::MenuPresenter.to_html` results in `NoMethodError` when calling `.homepage?`


kfrz:
When attempting to render a navigation with a given `Spina::MenuPresenter`, `ruby/3.1.3/lib/ruby/gems/3.1.0/gems/activemodel-7.0.3.1/lib/active_model/attribute_methods.rb:458` is called. This results in a `NoMethodError` when following the recommended [navigation guide](https://spinacms.com/guides/themes-content/adding-navigation-to-your-website). 

````

.../attribute_methods.rb:458:in `method_missing': undefined method `homepage?' for 
#<Spina::NavigationItem id: 36, page_id: 3, navigation_id: 3, position: 0, ancestry: nil, created_at: "2022-12-17 13:11:20.015606000 +0000", updated_at: "2022-12-17 13:11:20.015606000 +0000"> (NoMethodError)

````

Looks like the `homepage?` method is defined on the [`::Page` object class](https://github.com/SpinaCMS/Spina/blob/bce2cba8370471a0c75301f3442d8463780515f1/app/models/spina/page.rb#L61). Each `Navigation` has_many `Pages` :through `NavigationItems`. 

Mapping with `Navigation#pages.map(&:homepage?)` works - so it seems `NavigationItem#homepage?` is the culprit. Somewhere along the way `NavigationItems` has changed, but it's my first foray into this project's source and I would defer to more familiar devs on the potential path forward. My suggestion would be to delegate this `homepage?` method or wire up the `NavigationItems` class in such a way that it gracefully handles this error. 

https://github.com/SpinaCMS/Spina/blob/main/app/presenters/spina/menu_presenter.rb#L93

PR #1165, opened 2022-12-19
#1164 - Bugfix for` NoMethodError` when calling `.homepage?` on `NavigationItem` object

kfrz:
Fix NoMethodError resulting from checking the NavigationItem and not… the Page itself.

### Context

When attempting to render a navigation with a given Spina::MenuPresenter, ruby/3.1.3/lib/ruby/gems/3.1.0/gems/activemodel-7.0.3.1/lib/active_model/attribute_methods.rb:458 is called. This results in a NoMethodError when following the recommended [navigation guide](https://spinacms.com/guides/themes-content/adding-navigation-to-your-website).

More details in #1164 

This PR checks against the `Page` itself to see if it's the homepage; fixing an error. 

### Guidance to review
- [ ] ? Is the `Page` object the only type that would be included in the `NavigationItem` enumerable?

### DIFF
--- a/app/presenters/spina/menu_presenter.rb
+++ b/app/presenters/spina/menu_presenter.rb
@@ -91,7 +91,7 @@ module Spina
     end
 
     def parent_of_current?(item)
-      return false if item.homepage?
+      return false if item.page.homepage?
       Spina::Current.page.materialized_path.starts_with? item.materialized_path
     end
   end


kfrz:
I think I realize now this only works for menus built with `Navigation` objects and not a collection of `Page` objects, but would it make sense to check for which class `item` is before? Or just do a circuit like: if `item.page.homepage? || item.homepage?`

Bramjetten:
Fixed in #1169

PR #1180, opened 2023-01-17
Search in Media Library

noahsettersten:
### Context

We've had a client ask before about whether they can search within the media library to find uploaded items. I've experimented a bit with adding a simple search input in the header there to allow filtering by filename.

Here's a video demo of the UI for this PR: https://www.loom.com/share/52e45ba749404711b59593e0473b8b4a

In this pull request, I've added a search bar to the tabs within the Media Library at the top right. Typing text and pressing enter or clicking Search filters the list to only those media library items that contain the search text.

### Guidance to review

Using the feature is fairly straightforward:
1. Log in to the Spina admin panel.
2. Browse to the Media Library.
3. Enter a filename in the search box at the top-right and confirm that the list of files is filtered as expected.
4. Click the "X" icon to clear the search and restore the full list.

### Questions

- Does this feature make sense within Spina core? I can see it being helpful for finding existing library files, but those files are not always named clearly for searching.
- Any thoughts on the UI? I'm not sure what the best placement for the searching UI should be. Perhaps right-aligned beneath the action buttons could work well also.

I'm open to any suggestions or ideas 👍 If this is a helpful feature, I'd be open to adding the ability to search the main list of Spina pages as well.

### DIFF
--- a/app/components/spina/forms/search_component.html.erb
+++ b/app/components/spina/forms/search_component.html.erb
@@ -1,0 +1,16 @@
+<div class="flex items-center relative">
+  <div class="absolute left-3" style="left: 0.75rem;">
+    <%= helpers.heroicon('search', style: :solid, class: 'text-gray-300 w-5 h-5') %>
+  </div>
+
+  <%= @f.search_field @method, 
+    value: params[@method.to_sym],
+    class: "form-input font-medium block pl-10 pr-2 py-[0.3rem]"
+  %>
+
+  <div class="absolute right-2">
+    <%= link_to request.path do %>
+      <%= helpers.heroicon('x', style: :solid, class: 'text-gray-300 w-5 h-5') %>
+    <% end %>
+  </div>
+</div>
--- a/app/components/spina/forms/search_component.rb
+++ b/app/components/spina/forms/search_component.rb
@@ -1,0 +1,12 @@
+module Spina
+  module Forms
+    class SearchComponent < ApplicationComponent
+      attr_accessor :f, :method
+
+      def initialize(f, method)
+        @f = f
+        @method = method
+      end
+    end
+  end
+end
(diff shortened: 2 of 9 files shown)


noahsettersten:
I added a basic system test for this change, but I was unable to get the system tests working in my local development environment. I manually added a `Spina::Account` record to the test database set to the `"demo"` theme as that appeared to be necessary here.

Instead of completing successfully, the system tests hang for some time before throwing a `Net::ReadTimeout` error. This appears to happen visiting the initial login page on line 11 of `application_system_test_case.rb`. I wonder whether the test server is not responding for some reason. I'm able to run `RAILS_ENV bundle exec rails server` without issue outside of the test suite after creating the `Spina::Account` record though.

Outside of the system test, I'm open to other ideas for adding tests as well. I could perhaps create an integration test in place of this system test, if that's preferred.

Bramjetten:
I really like this!

Couple of notes:
- Would love to also add this to the modal UI when selecting images elsewhere
- I think we can move the has_one_attached method and with_filename scope into a concern 
- I feel searching should happen instantly, without having to submit the form manually
- The search input already includes a native "x"-button in Safari for me, there's two overlapping now

Cool if I add commits to your branch?

Bramjetten:
I played around with your branch and sprinkled some Hotwire on top:

https://user-images.githubusercontent.com/423116/216839055-ffcbdece-0e69-45c7-afd4-406a41bdcdbe.mov

Bramjetten:
Another sketch with a subtle UI:

https://user-images.githubusercontent.com/423116/216839501-9ccad3fe-1f28-4af9-ab11-16f7db263418.mov

Bramjetten:
Maybe it's better to put it here instead though:
CleanShot 2023-02-05 at 20 12 11@2x

Bramjetten:
CleanShot 2023-02-07 at 10 06 15@2x


CleanShot 2023-02-07 at 10 06 31@2x

Bramjetten:
https://user-images.githubusercontent.com/423116/217218220-05090980-6ac6-4f4f-abbb-175604b6fd8b.mov

noahsettersten:
@Bramjetten You've taken this in a great direction; I appreciate the ideas and work you built further! Since you've opened a new PR to keep this moving, I'll close out this one in favor of your newer work in #1192.

PR #1192, opened 2023-02-07
Search

Bramjetten:
Based on @noahsettersten's work in #1180 

CleanShot 2023-02-07 at 11 40 30@2x

https://user-images.githubusercontent.com/423116/217222545-9d115fed-420f-4c49-8bf0-8daf6b4467a7.mov

### DIFF
--- a/app/components/spina/forms/search_component.html.erb
+++ b/app/components/spina/forms/search_component.html.erb
@@ -1,16 +1,12 @@
 <div class="flex items-center relative">
-  <div class="absolute left-3" style="left: 0.75rem;">
+  <div class="absolute left-2">
     <%= helpers.heroicon('search', style: :solid, class: 'text-gray-300 w-5 h-5') %>
   </div>
 
-  <%= @f.search_field @method, 
+  <%= f.search_field @method, 
     value: params[@method.to_sym],
-    class: "form-input font-medium block pl-10 pr-2 py-[0.3rem]"
+    placeholder: t("spina.search"),
+    class: "form-input rounded-full pl-8 text-sm h-9 w-full",
+    data: {action: "input->form#requestSubmit"}
   %>
-
-  <div class="absolute right-2">
-    <%= link_to request.path do %>
-      <%= helpers.heroicon('x', style: :solid, class: 'text-gray-300 w-5 h-5') %>
-    <% end %>
-  </div>
 </div>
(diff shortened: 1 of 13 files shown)


app/controllers/spina/admin/attachments_controller.rb
```diff
@@ -4,11 +4,7 @@ class AttachmentsController < AdminController
       before_action :set_breadcrumbs
 
       def index
-        @attachments = Attachment.sorted.with_attached_file.page(params[:page]).per(25)
-
-        if params[:query].present?
-          @attachments = @attachments.with_filename(params[:query])
-        end
+        @attachments = Attachment.sorted.with_attached_file.with_filename(params[:query].to_s).page(params[:page]).per(25)
```
noahsettersten, line 7: This is much cleaner than the conditional I had on the query param 👍

app/components/spina/forms/search_component.html.erb
```diff
@@ -1,16 +1,12 @@
...
     <%= helpers.heroicon('search', style: :solid, class: 'text-gray-300 w-5 h-5') %>
   </div>
 
-  <%= @f.search_field @method, 
+  <%= f.search_field @method, 
     value: params[@method.to_sym],
-    class: "form-input font-medium block pl-10 pr-2 py-[0.3rem]"
+    placeholder: t("spina.search"),
+    class: "form-input rounded-full pl-8 text-sm h-9 w-full",
+    data: {action: "input->form#requestSubmit"}
```
noahsettersten, line 10: The search on input is a nice touch. 👌

app/components/spina/forms/search_component.html.erb
```diff
@@ -1,16 +1,12 @@
...
   %>
-
-  <div class="absolute right-2">
-    <%= link_to request.path do %>
-      <%= helpers.heroicon('x', style: :solid, class: 'text-gray-300 w-5 h-5') %>
```
noahsettersten: My initial development was in Firefox, and I didn't consider that Safari added its own close button. It makes sense to leave the management of the field to the browser.

noahsettersten (commented):
This is great! I pulled it into an app and explored a bit, and it works quite smoothly and unobtrusively. I did run `bundle exec rails spina:tailwind:build` before the search field displayed correctly. That may be a side effect of this app, but I thought I'd mention it.

I'll close my PR in favor of the expanded work you built here, Bram. Thanks for diving into it!

app/components/spina/forms/search_component.html.erb
Bramjetten: I was working on another search input for the new PageLink part which includes a debounce method. When that's merged, this will also debounce so it's not firing search requests like crazy as it does now.

noahsettersten:
@Bramjetten I've updated to the new release of Spina and this is working great 👍 I appreciate you building this in!

PR #1200, opened 2023-02-14
Collapse child pages

Bramjetten:
This PR adds the ability to collapse and expand child pages in the admin UI. Additionally, this PR adds a db migration that adds two cache columns for ancestry, making pages#index waaay faster.

CleanShot 2023-02-14 at 15 44 49@2x

Child pages are collapsed by default for resources/page collections.

Sidenote: isn't it great this PR includes *zero* changes to controllers?!


mattpolito:
The changes you made work. The only weirdness I'm having is style related. I'm not seeing the nested styling like you have in the screenshots and placeholder div is not hidden when collapsed.
Screenshot 2023-02-15 at 9 51 37 PM
Screenshot 2023-02-15 at 9 52 05 PM
Screenshot 2023-02-15 at 9 52 15 PM

Bramjetten:
Both of those things are by design actually, haha.

The background color is the hover-state. That small line *is* the placeholder, which I show when pages are collapsed to indicate there are pages under there. I can see in your example that it looks a bit weird, so perhaps it’s better to drop that placeholder altogether.

mattpolito:
I feel like the chevron is enough of an indicator that there are additional pages.

Also in previous design examples the child pages were indented when opened. That is not what I’m experiencing anymore once the turbo data request was implemented.

Bramjetten:
I deleted the placeholder background.

Your indentation is probably missing due to the new db migration. If you re-save your pages I think the indentation should be fixed.

mattpolito:
> Your indentation is probably missing due to the new db migration. If you re-save your pages I think the indentation should be fixed.

Resaving all the pages fixed indentation. (This should probably be notated somewhere in the changelog if it's going to be necessary)

Screenshot 2023-02-21 at 9 57 49 AM 2

And the removal of the placeholder makes everything look as I would expect.

This PR looks good to go!

PR #1214, opened 2023-03-16
Rewrite rules change

Jonathan-Jacob:
Change the rewrite rules method so that the new_path is always the currently active one. This way, there will always be only one redirect instead of multiple ones in case a page handle gets renamed more than once.

Also, this gets rid of the following problem that exists until now:

1. You start with page A that has path a, and page B that has path b.
2. You set a new path c for page A, that triggers a rewrite rule from a to c.
3. You set a new path d for page A, that triggers a rewrite rule from c to d.
4. You set a new path c (the same one you set for page A in step 2) for page B, that triggers a rewrite rule from b to c.
5. You visit path a where page A has been and will get redirected to path c, where page B is now – **that's not what you want.**
You visit path b where page B has been and will get redirected to path c where page B is now – that's what you want.
You visit path c where page A has been but page B is now, so you will not get redirected – this is probably also what you want.

With this PR, you will now have the following:

1. You start with page A that has path a, and page B that has path b.
2. You set a new path c for page A, that triggers a rewrite rule from a to c.
3. You set a new path d for page A, that triggers a rewrite rule from c to d and also change the rule set in step 2 – this rewrite rule will now be from a to d.
4. You set a new path c (the same one you set for page A in step 2) for page B, that triggers a rewrite rule from b to c and will also remove the rewrite rule from c to d – so you have 2 rewrite rules now, from a to d and from b to c.
5. You visit path a where page A has been and will get redirected to path d, where page A is now – **that's what you want.**
You visit path b where page B has been and will get redirected to path c where page B is now – that's also what you want.
You visit path c where page A has been but page B is now, so you will not get redirected – this is probably also what you want.

### DIFF
--- a/app/models/spina/page.rb
+++ b/app/models/spina/page.rb
@@ -115,9 +115,11 @@ module Spina
     end
 
     def rewrite_rule
-      RewriteRule.where(old_path: materialized_path).delete_all
-      RewriteRule.where(new_path: old_path).update(new_path: materialized_path)
-      RewriteRule.where(old_path: old_path).first_or_create.update(new_path: materialized_path) if old_path != materialized_path
+      if old_path != materialized_path
+        RewriteRule.where(old_path: materialized_path).delete_all
+        RewriteRule.where(new_path: old_path).update(new_path: materialized_path)
+        RewriteRule.where(old_path: old_path).first_or_create.update(new_path: materialized_path)
+      end
     end
 
     def localized_materialized_path
--- a/app/models/spina/page.rb
+++ b/app/models/spina/page.rb
@@ -39,7 +39,7 @@ module Spina
     after_save :touch_navigations
 
     # Create a 301 redirect if materialized_path changed
-    after_update :rewrite_rule
+    after_update :redirect_to_new_path, if: :path_changed?
 
     before_validation :set_materialized_path
     validates :title, presence: true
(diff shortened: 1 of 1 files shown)


Bramjetten:
I really need to draw this out for myself to grasp it fully, haha. Perhaps it'd be easier to add the if statement to the after_update call:
`after_update :create_rewrite_rule, if: -> { old_path != materialized_path }`

Bramjetten:
Perhaps we could change a `def path_changed?` method, to make it more clear.

Jonathan-Jacob:
> I really need to draw this out for myself to grasp it fully, haha. Perhaps it'd make more sense to add the if statement to the after_update call: `after_update :create_rewrite_rule, if: -> { old_path != materialized_path }`

Yeah that certainly makes sense!

Jonathan-Jacob:
> Perhaps we could add a `def path_changed?` method, to make it more clear. That way you could have: `after_update :redirect_to_new_path, if: :path_changed?`
> 
> This makes it much clearer to me.

So that `def path_changed?`method would basically  check whether `old_path != materialized_path`? Yeah makes sense

Jonathan-Jacob:
> Perhaps we could add a `def path_changed?` method, to make it more clear. That way you could have: `after_update :redirect_to_new_path, if: :path_changed?`
> 
> This makes it much clearer to me.

I implemented that in a new commit, please have a look

Bramjetten:
Thank you for making those changes. I’m trying to come up with a way to clean up that method now. Would it make sense to move creating/deleting those rewrite rules to a bg job?

Jonathan-Jacob:
Do you think this will take up too much runtime otherwise? I'm wondering why that check has failed – it passed before and all I did was refactor, so my guess is that this is some kind of a false positive. My proposed change would replace 1 ActiveRecord query with 3 queries, so I'd be quite surprised if this really took more than a couple of milliseconds.

But this is all not more than my best semi-educated guess. Do you have a way to compare runtimes?

Issue #1265, opened 2023-07-16
Improve display of large images inserted into Rich Text (Trix) content


wakproductions:
When selecting a large image from the media picker, it overflows the page. See this example here:

![image](https://github.com/SpinaCMS/Spina/assets/1912237/1f702d08-1dff-41da-8cd5-db57c8116faf)
![image](https://github.com/SpinaCMS/Spina/assets/1912237/ae7ea597-7aea-4fba-8bcf-6bf96190ecc0)

### Proposed Enhancement

A drop down that would allow the end user to pick a smaller variant that would be inserted into Trix. Default setting would be "Original Size". If an option other than "Original Size" is selected, the variant image displayed will be clickable, navigating the user to the original size image.

![image](https://github.com/SpinaCMS/Spina/assets/1912237/94f20999-0685-45d3-9e25-1e8cf0cef92f)

### Additional Notes

Image styling can be tweaked using CSS (`max-width`) on the application end. However, a very large image could have slow loading performance. A hyperlink that opens the full size image in the browser could also be useful for images where the viewer may want to see the high resolution up close for more detail.

wakproductions:
I'm working on a PR to implement this, however having trouble deciding when the selected variant URL should be corrected. The media picker works almost entirely on the frontend using the metadata attached to the Image component (`image_component.html.erb`). But I think trying to populate the variant URLs in this metadata at the time the media picker loads would also prompt generation of all the image variants at this point, which may be undesirable. 

I think the best time to generate the variant would be after the user selects the variant value, which would be either when the selection in the "variant" box changes, or during processing of the `media-picker:done` event. Either case would require a call to the backend to retrieve the variant URL.

It would be helpful to have some feedback on this idea. Thanks for making an overall terrific CMS program @Bramjetten!

wakproductions:
After thinking about this, I'm not sure that the option to select a variant size is necessary. Using the "preview_url", which does a `resize_to_limit` of 1600x1200 I think would serve most cases. Here's what my example above looks like at that scale, with a `max-width: 90%` set on the stylesheet within the application running it. I think that sizing works just fine for large images on a blog. The important thing would be making a link that expands to the full image when the user clicks on it.

![image](https://github.com/SpinaCMS/Spina/assets/1912237/6485d4e8-2998-4dcf-8b96-2ec00da84f19)

[...]

PR #1266, opened 2023-07-17
Link embedded rich text images to full size image

wakproductions:
### Context
See discussion: https://github.com/SpinaCMS/Spina/issues/1265

Would be nice if clicking on an embedded image takes the user to the original, full sized image. This would be useful for images with details that are best viewed at high resolution up close, or charts and tables that are difficult to read when downsized.

The proposed enhancement matches a similar feature found in Wordpress. Here is an example of a page where images are clickable to see the full sized image: https://winningwp.com/godaddy-vs-siteground-which-is-best-for-wordpress-hosting/

Wraps images in an `<a>` tag with a link to the original image.

### Guidance to review

Up to @Bramjetten to decide whether this is a beneficial change.


wakproductions:
I'm having trouble running the test suite locally. For some reason, when I run `rake test` it truncates the `schema_migrations` table and then thinks it has migrations to run. 🤷‍♂️

Bramjetten:
Thanks for creating this PR! 

By default, the images Spina inserts are resized to 2000x2000, *not* the original size. You can override that by setting `embedded_image_size`.

app/assets/javascripts/spina/controllers/auto_file_upload_controller.js
```diff
@@ -36,14 +36,4 @@ export default class extends Controller {
...
-  #imageData(imageMeta) {
-    return {
-      filename: imageMeta.dataset.filename,
-      signedBlobId: imageMeta.dataset.signedBlobId,
-      imageId: imageMeta.dataset.imageId,
-      embeddedUrl: imageMeta.dataset.embeddedUrl,
-      thumbnail: imageMeta.dataset.thumbnail
-    }
-  }
-
```
wakproductions, line 48: Accidentally forgot to delete this orphaned code in my last PR 🫢

app/assets/javascripts/spina/controllers/trix_controller.js
```diff
@@ -33,9 +33,12 @@ export default class extends Controller {
...
   insertAttachment(event) {
-    let attachment = new Trix.Attachment({content: `<span class="trix-attachment-spina-image" data-label="Alt text">
-      <img src="${event.detail.embeddedUrl}" />
-    </span>`, contentType: "Spina::Image"})
+    const content =
+      `<span class="trix-attachment-spina-image" data-label="Alt text">
+       <a href="${event.detail.originalUrl}"><img src="${event.detail.embeddedUrl}" /></a>
+       </span>`
+
+    let attachment = new Trix.Attachment({content, contentType: "Spina::Image"})
```
wakproductions, line 41: This just adds a link wrapper to images allowing the image to be clickable, leading to the full sized image.

wakproductions:
Just added a small update to the code. Here's why this feature is useful. For things like stock charts or other kinds of data, the user might want the high rez version of the image to better see details.

spina-click-to-see-full-image-demo

wakproductions:
> By default, the images Spina inserts are resized to 2000x2000, _not_ the original size. You can override that by setting `embedded_image_size`.

Thanks I made that correction in the docs addition here.

wakproductions:
@Bramjetten if you don't like this feature, what do you think about modifying `RichTextPresenter` so that image type of data is passed off to another object, maybe named `Embeds::Image`. That object could be overridden in the main program to have whatever behavior the website creator wants for images? We can even create configuration settings in `initializers/spina.rb` to be accessed by that object so that rich text images can be handled however specified?

MSchmidt:
IMHO I wouldn't want this as a default behaviour. Or if default at least configurable. I'd prefer to see images that can link to anything but spina's trix doesn't allow links around images right now.

wakproductions:
> IMHO I wouldn't want this as a default behaviour. Or if default at least configurable. I'd prefer to see images that can link to anything but spina's trix doesn't allow links around images right now.

Understandable. I might try try a different approach with this then - making an image embed component that has configurable properties like "clickable" and "caption" that would have its own wrapper template.

PR #1269, opened 2023-07-24
Custom url navigations

vanillaHafer:
### Context

Currently, Spina::NavigationItems can only link to other main Spina pages and they cannot be edited. It would be nice if Custom URLs were able to be added in, to allow more customization to the user.

The 2 main changes in this PR are:

1. Being able to edit existing navigation items
image
2. Being able to use custom URLs with their labels, instead of having to select from the list of pages
image

### Guidance to review

Everything existing functions as normal. All changes are simply additions.


app/models/spina/navigation_item.rb
```diff
@@ -25,6 +25,14 @@ def menu_title
       end
     end
 
+    def materialized_path
```
Bramjetten, line 28: This could be rewritten as:
```ruby
def materialized_path
  page&.materialized_path || url
end
```

or like this if you prefer something a bit more explicit:

```ruby
def materialized_path
  return page.materialized_path if page
  url
end
```

app/models/spina/navigation_item.rb
```diff
@@ -11,8 +11,40 @@ class NavigationItem < ApplicationRecord
     scope :in_menu, -> { joins(:page).where(spina_pages: {show_in_menu: true}) }
     scope :active, -> { joins(:page).where(spina_pages: {active: true}) }
 
-    validates :page, uniqueness: {scope: :navigation}
+    validates :page, uniqueness: {scope: :navigation, allow_nil: true}
+    validate :url_and_url_label_presence
+    validate :url_or_page_presence
 
-    delegate :menu_title, :materialized_path, :draft?, :homepage?, to: :page
+    delegate :draft?, :homepage?, to: :page
```
Bramjetten, line 18: I think these also need their own method now, defaulting to false when there's no page. We might be able to get away with `allow_nil: true` here.

test/dummy/db/schema.rb
```diff
@@ -124,12 +124,14 @@
...
   create_table "spina_navigation_items", id: :serial, force: :cascade do |t|
-    t.integer "page_id", null: false
+    t.integer "page_id"
     t.integer "navigation_id", null: false
     t.integer "position", default: 0, null: false
     t.string "ancestry"
     t.datetime "created_at", precision: nil
     t.datetime "updated_at", precision: nil
+    t.string "url"
+    t.string "url_label"
```
Bramjetten, line 134: Why not `menu_title` like for pages? How do you handle translations?

Bramjetten (commented):
Thanks for working on this! I like the idea here, but I'm not sure about the UI implementation yet. Few things to consider:
- How do you handle translations? This goes for the label, but also the URL
- Instead of having a button "Add menu item", perhaps it should be two buttons: "Add page" and "Add URL" which will then have their own forms.

mattpolito:
> * Instead of having a button "Add menu item", perhaps it should be two buttons: "Add page" and "Add URL" which will then have their own forms.

@Bramjetten I guess it depends on if we have that design language elsewhere in the site. Personally I would think that I'm going to this section to add a navigation item... so already the language of "Add menu item" differs but at least it's a singular thing. If we add choices here, I would think that someone would always choose "Add URL" as it makes the most sense for a nav item. 

For confusion sake, I would suggest that there be a single link to open a form where you'd then choose the kind of thing you want to create. In the end you're always creating "A menu item".

mattpolito:
@Bramjetten Also we noticed that a test failure was introduced by a previous commit so you'll see the fix for that in here as well.

test/dummy/db/schema.rb
mattpolito: We ended up making changes so there is a `url_title` attribute that get translated through mobility.

app/models/spina/navigation_item.rb
vanillaHafer: Great suggestion. I went ahead and made those corrections.

app/models/spina/navigation_item.rb
vanillaHafer: Good call! That was an oversight on my part. Went ahead and added the `allow_nil: true` to take care of this.

Bramjetten:
@mattpolito I've noticed some flakiness with a certain Resource test, so I might have to fix that anyway.

mattpolito:
@Bramjetten do you have any additional feedback on this one?

Bramjetten:
I would very much like to have this feature, so let's work on getting it in Spina!

Couple of things I notice in the UI in no particular order:
- The "Choose page" select is a simple dropdown which is hard to use when you have many pages. Maybe we can borrow PageLink's select with search functionality.
- Using a toggle to switch forms feels odd. Also, the window jumps when you do that.
- The input fields don't have autofocus
- Perhaps instead of a toggle, it can simply be a link to a different form
- There's no validation on adding a URL
- It doesn't support translations
- If you can change the label for custom URLs, shouldn't you be able to do that for pages as well?
- I feel the navigation items themselves should have a little indicator to tell you it's a custom URL and not a page

Couple of things I notice in the code:
- Why not use/update the SwitchComponent?
- Can we have one migration instead of two?
- Instead of altering `navigation_item_params` in the controller, I'd prefer to use Turbo Frames to change the form instead of showing/hiding input fields with JS.

mattpolito:
> * The "Choose page" select is a simple dropdown which is hard to use when you have many pages. Maybe we can borrow PageLink's select with search functionality.

I completely agree with this needing improvement. However it is not part of adding a custom url so I'd leave it for another PR.

> * Using a toggle to switch forms feels odd. Also, the window jumps when you do that.

Clearly you haven't been a fan of the toggle and I don't think a separate button is the right choice either. How do you feel about that model opening with tabs? One for "Page" and another for "URL"?

> * The input fields don't have autofocus

This can be easily addressed

> * Perhaps instead of a toggle, it can simply be a link to a different form

See above comment regarding toggle/form

> * There's no validation on adding a URL

Technically there is a validation. At least one that says it has to be populated if you select the custom url option. Are you referring to something more specific validating the URL itself? Also due to how modals are implemented currently, It does not allow for errors to come back from turbo. We've been wanting to address this anyway. Again, I think that'd be change but not part of this PR.

> * It doesn't support translations

We added translations the other day. Did you not see that change or are you referring to something else?

> * If you can change the label for custom URLs, shouldn't you be able to do that for pages as well?

We didn't make that change because currently you have the mechanism to change a page label on the page itself. Not sure if we want to immediately tackle this but we think it should actually live here in the navigation item. That way we get a uniform experience as well as the ability to have have pages in multiple navigations that can have a different name than the singular one defined on the page.

> * I feel the navigation items themselves should have a little indicator to tell you it's a custom URL and not a page

I agree with this.

> * Why not use/update the SwitchComponent?

If you don't like the toggle, then there's no reason to use the switch component. So we just need to make a decision on UX.

> * Can we have one migration instead of two?

We *CAN* however we're addressing two different tables so it seems appropriate to have two different migrations. With that said... Is there a reason we're using integer migration versions vs timestamp?

> * Instead of altering `navigation_item_params` in the controller, I'd prefer to use Turbo Frames to change the form instead of showing/hiding input fields with JS.

This again refers to the decision that needs to be made about the UX.

Overall... we can get the changes in that we all agree on and let's discuss how you feel about the form UX so we can move forward with the other changes.

vanillaHafer:
We made a few changes recently based on your feedback. We got away from the toggle mechanism and have two separate forms using tabs:
image
image

We also added in the indicator showing if a navigation item is an external link:
image

We also addressed validating the url and and url_title in the form:
image

We also implemented autofocus for the url form.

Whenever you get a chance, feel free to look over the changes and let us know if there's any other feedback you have!

Bramjetten:
Very cool, nice!

I think we can improve the UI a bit more. I was thinking about something like this:
CleanShot 2023-07-31 at 16 27 46@2x

This way, adding a page to the menu is still the primary way of adding items while simultaneously making it very clear what the other option is meant for.

As for the indicator on the item itself, perhaps we can use this icon with a small tooltip: 
CleanShot 2023-07-31 at 16 30 20@2x

Excuse me if I missed it, but have you taken a look at adding something for translations?

vanillaHafer:
I went ahead and implemented those UI changes, so hopefully they are looking more like what you were expecting.

### Removed tabs in favor of links
image
image

### Changed indicator
image

### Translations

As far as translations go, we're handling translations for the UI elements the same as all the other elements, by adding to the `.yml` files under `config/locales/`. And then regarding actual URL Title translations we are using a translation table for navigation items. As far as we understand, that's enough to handle translations, but please let us know if we are missing something.

Bramjetten:
This is great! 

What I mean with the translations is that I don't think the UI currently supports editing multiple translations. How can end-users add multiple translations for a menu item?

mattpolito:
> How can end-users add multiple translations for a menu item?

How is that working anywhere else? I don't recall seeing that functionality.

Bramjetten:
> How is that working anywhere else? I don't recall seeing that functionality.

CleanShot 2023-07-31 at 21 23 05@2x

vanillaHafer:
Since we don't use translations for our application, the translation feature was a bit foreign to us. Thanks for being patient with us as we learn more about Spina to get this feature rolling.

Looking at the current implementation I was able to throw something together for the Navigations/Navigation Items.

As you can see on the Navigations screen we now have a Translation Component in the upper right:
image

Which will display Navigation Items using their correct translations:
image

When adding in a navigation, we can choose which translation we are adding:
image

After adding in a navigation item and giving it, say, an English translation first, we can now see that some translations are missing url_titles:
image

Which is also seen when editing that particular navigation item:
image

As usual, please give us any feedback you have on this so we can correct anything you think should change!

Bramjetten:
I think we can improve the UI for switching locales. I think it’s best if we make it possible to fill in all translations in one go, as that’s probably the most common workflow. Perhaps for now it’s best to simply list the different locale fields from top to bottom and improve the UI further in a different PR. That way we can get this merged now and enable it for everyone :)

mattpolito:
>  I think it’s best if we make it possible to fill in all translations in one go, as that’s probably the most common workflow.

@Bramjetten is there an example of this workflow currently in the codebase?

Bramjetten:
No there isn’t, but the easiest way is to simply loop over all locales in Spina.config.locales and render the fields manually.

mattpolito:
Since that is a new UI concept, I would really suggest it be separate from this PR. The data flow wouldn't change so doing it at a later time would not make a difference.

Bramjetten:
@mattpolito Agreed that we can merge this into the main branch now and clean up the UI in a separate PR. I’ll create an issue for it.

Bramjetten:
@mattpolito Do you know why the tests are failing?

mattpolito:
@Bramjetten thats an interesting failure. I'll look at it or have @vanillaHafer double check it. It's failing on Minitest so it may be a CI issue.

mattpolito:
@Bramjetten suite on this branch was passing. The error that is currently happening seems to have been introduced from #1273

Bramjetten (approved):

PR #1272, opened 2023-07-25
Fix account test

vanillaHafer:
### Context

On a recent PR for adding ancestry into custom_pages, there was a failing test.

These changes are to fix that test, and add a default setting for the Page factory

### DIFF
--- a/test/factories/pages.rb
+++ b/test/factories/pages.rb
@@ -1,5 +1,6 @@
 FactoryBot.define do
   factory :page, class: "Spina::Page" do
+    sequence(:title) { |n| "Page #{n}" }
     draft { false }
     active { true }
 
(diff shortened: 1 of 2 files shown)


Bramjetten (approved):

Bramjetten:
It still fails?

mattpolito:
@Bramjetten That's a webdriver failure. You may want to force a re-run.

vanillaHafer:
The "flicker" around the first test failing was due to comparing Page objects and their json_attributes not being guaranteed to be in the same order. I handled that by just comparing object ids.

The 2 most recent test fails are due to the web driver issue.

Bramjetten:
I'm also seeing intermittent web driver issues on other projects. Looks like something changed over there.

PR #1279, opened 2023-07-30
Use default theme by default

sedubois:
### Context
The generator provides defaults for all options except for the theme.

Provide a default theme value so that installation can proceed simply by pressing Enter.

### Guidance to review
run `rails g spina:install` and observe that the default theme is used when just typing Enter.


mattpolito:
👍

Bramjetten (approved):

PR #1291, opened 2023-08-21
Custom url navigations – second draft

Bramjetten:


mattpolito:
@Bramjetten Seems fine to me. Looks like you removed translations?

PR #1297, opened 2023-08-26
Use left_outer_joins for NavigationItem scopes - Fixes #1294

### LINKED ISSUE #1294: Navigation Items with URL kind not presented by MenuPresenter
Currently the MenuPresenter#scoped_collection method applies a number of scopes to a navigation's items, most of which join to the Pages class, causing the URL kind navigation items to not return to be presented.

Bramjetten:
Because of the addition of custom URLs in NavigationItems, scopes on NavigationItem should use `left_outer_joins(:page)` instead of `joins(:page)`.

Also, it doesn't make sense to use the `regular_pages` scope in the MenuPresenter.


Bramjetten:
@aseroff Can you check if this fixes your issue?

aseroff:
Hole in one again

PR #1380, opened 2024-08-01
Add a with_library_path scope to Attachable

MSchmidt:
### Context
I'm working on markdown support for Spina. To show images from the library I want them to be included with this easy syntax: `Alt` within the markdown.

This scope on Attachable helps to fetch the right images based on that syntax.

### Guidance to review
Right now Spina doesn't use this scope internally. So I can see how this might not be accepted at the time. Also it's still using `ILIKE` because that's what the `main` branch does. Will change it once you merge the sql PR.

### DIFF
--- a/app/models/concerns/spina/attachable.rb
+++ b/app/models/concerns/spina/attachable.rb
@@ -11,6 +11,25 @@ module Spina
           "%" + Image.sanitize_sql_like(query) + "%"
         )
       end
+
+      scope :with_library_path, ->(query) do
+        folder_name, file_name = query.split("/")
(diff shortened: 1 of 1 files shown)


Bramjetten:
That's really really cool! Are you also looking at including a Markdown-editor (with preview?)

Bramjetten:
Not stale!

how would House (MD) fit into this?

MSchmidt:
House is not released yet, but I guess as trix alternative for the MD Parts it would work great. We'd need to expend it with an spina image picker though. This PR is only the preparation for that.

PR #1387, opened 2024-10-21
fix: Use Tailwindcss::Ruby executable

mattpolito:
### Context
`Tailwindcss::Engine.root` is not the location of the executable anymore. We can utilize `Tailwindcss::Ruby.executable` the same way th

We can utilize `Tailwindcss::Ruby.executable` the same way that [`tailwindcss-rails`](https://github.com/rails/tailwindcss-rails) does.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,7 +18,7 @@ PATH
       rails (>= 6.0)
       sprockets-rails
       stimulus-rails (>= 0.7.0)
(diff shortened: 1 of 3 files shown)


Bramjetten:
Nice! Maybe also drop the tailwindcss-rails dependency?

mattpolito:
Were you keeping it around previously just for that command?

Bramjetten:
Yes

PR #1391, opened 2024-11-02
Feature translate locales with region

pedrocarmona:
### Context

Allow locales with region (example pt-PT: Portuguese from Portugal region)

- Creates content methods that follow mobility convention (underscore locales with region)
- Adds Portuguese with Portugal region - there was already Portuguese with Brazilian region)

### Guidance to review

```
# config/initializers/spina.rb
Spina.configure do |config|
  ..
  config.locales = [ :'pt-PT', :en ]
```

```
# config/initializers/mobility.rb
Mobility.configure do
  ... 
  locale_accessors [ :'pt-PT', :en ]
```

```
# config/application.rb
  ...
  config.i18n.default_locale = :'pt-PT'
```

### DIFF
--- a/app/models/concerns/spina/translated_content.rb
+++ b/app/models/concerns/spina/translated_content.rb
@@ -7,14 +7,14 @@ module Spina
     included do
       # Store each locale's content in [locale]_content as an array of parts
       Spina.locales.each do |locale|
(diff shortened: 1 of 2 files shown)


pedrocarmona:
maybe this is not the best approach, maybe for now should be enough to convert `pt-BR.yml` into `pt.yml`

or even use i18n fallbacks

PR #1409, opened 2025-10-14
Allow multiple dialogs by using the default dialog mechanism in trix

simmerz:
Slight design change in the link modal to stop it overlapping other editors.

Primary focus of this PR is to allow multiple buttons to have dialogs by reverting back to the native mechanism in trix.

I've had to add a comment with some tw classes, to force the tailwind builder to see them. Happy to remove, but didn't know a better way to handle that whilst still on older versions of tailwind-ruby/rails and not using the new engine features in importmaps.


Bramjetten:
Great!

I believe the original `RevealController` was also written by you, no?

Bramjetten:
Continued in #1418

Issue #1428, opened 2026-05-05
New feature idea: StreamField


wakproductions:
Sometimes it can be helpful to have an open-ended way of writing your content. You can use the Text page part which gives you rich text with possibility for custom components, but it can be difficult to recycle elements and sometimes the Trix editor doesn't work so well. It's also difficult to programmatically digest the rich text kind of content if you want to do certain mass edits.

Wagtail has a component called StreamField, a page part that allows you to mix and match subcomponents. It's great for writing blogs with varied content. I created a working prototype of this that can be used in Spina. It's based on the Repeater component, but allows you to choose from any set of components you pre-registered in the theme settings.

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/40f03955-0cda-48bf-95a9-0b3cab8637c7" />

Would you mind if I make a PR to integrate this component into core Spina? I can probably add documentation on how to use it wherever that exists as well.

[Youtube: StreamField in Wagtail](https://www.youtube.com/watch?v=9YLBJC1rPnk)

wakproductions:
@Bramjetten is this project still under active feature development?

Bramjetten:
> @Bramjetten is this project still under active feature development?

Yes! The feature you built doesn't match the philosophy I built Spina with. I think it's best to add this feature as a separate gem for now!

Also, I'm planning to merge Spina Pro's features with Spina, so we can have built-in datetime fields that you can order pages with as well.

wakproductions:
Thanks for responding. I understand if it's hard to find the time to look at all the new PRs. Let us all know if you need help with maintainers since this is a great Rubygem.

wakproductions:
> Yes! The feature you built doesn't match the philosophy I built Spina with. I think it's best to add this feature as a separate gem for now!

It would be helpful if you can elaborate how it does not fit into the philosophy. I can give this a try as a separate gem.

wakproductions:
@Bramjetten Streamfield is live!

https://github.com/wakproductions/spina-streamfield

macsimacsi:
I love this idea!

PR #1430, opened 2026-05-10
Feature: Markdown page part

wakproductions:
### Context

Some users may prefer to write formatting in Markdown as an alternative to the Trix editor formatting.

Add a Markdown page part to be included with Spina's default set of available page parts.

### Guidance to review

For including this in your Spina pages:
```ruby
# config/initializers/themes/default.rb
  theme.parts = [
    {name: "markdown", title: "Markdown section", part_type: "Spina::Parts::Markdown"},
  ]

  theme.view_templates = [
    {name: "page_parts_demo", title: "Page Parts Demo", parts: %w[markdown]}
  ]

# views/default/pages/page_parts_demo.html.erb
<h1><%= current_page.title %></h1>
<div class="pt-5">
  <%= content.markdown(:markdown) %>
</div>
```

It uses Kramdown as the Markdown rendering engine. If requested, I can add a setting that would allow the user to choose a different Markdown processor. Such a setting would allow even greater flexibility.

spina-markdown-component

### DIFF
--- a/app/models/spina/parts/markdown.rb
+++ b/app/models/spina/parts/markdown.rb
@@ -1,0 +1,11 @@
+module Spina
+  module Parts
+    class Markdown < Base
+      attr_json :content, :string, default: ""
+
+      def to_html
+        Kramdown::Document.new(content.to_s).to_html.html_safe
+      end
+    end
+  end
+end
--- a/app/presenters/spina/content_presenter.rb
+++ b/app/presenters/spina/content_presenter.rb
@@ -17,6 +17,10 @@ module Spina
       RichTextPresenter.new(view_context, html)
     end
 
+    def markdown(name)
+      Kramdown::Document.new(find_part(name)&.content.to_s).to_html.html_safe
+    end
+
     def image_tag(image, variant_options = {}, options = {})
       image = find_part(image) unless image.is_a? Spina::Parts::Image
       image_tag_options = {alt: image&.alt}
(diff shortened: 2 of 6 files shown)


Bramjetten:
Wouldn't it be preferred to instead switch from Trix to Lexxy? https://github.com/basecamp/lexxy
Lexxy supports Markdown out of the box.

You could already do what you're doing now with a regular MultiLine part.

wakproductions:
> Wouldn't it be preferred to instead switch from Trix to Lexxy? https://github.com/basecamp/lexxy Lexxy supports Markdown out of the box.

This is the first time I've heard of Lexxy. Looks like the project was only started last year and is intended to replace Trix. That might require some significant overhaul work on the admin editor, especially since it's configured to use a custom toolbar.

The reason I'm interested in directly using Markdown is to get more control over how the content is stored in the database. I don't like the limited options Trix editor allows for embedding media. See #1428 Streamfield for my proposed solution.

> You could already do what you're doing now with a regular MultiLine part.

This did not occur to me - just using a regular MultiLine part and interpreting it as markdown in the page rendering template. It wouldn't have a built in presenter though.

I gave it a try with a custom helper and this works:

```ruby
module ApplicationHelper
  def markdown(text_page_part)
    Kramdown::Document.new(text_page_part.to_s).to_html.html_safe
  end
end

# in page_parts_demo.html.erb (rendering template for the page)

<%= markdown(content.text(:simpletext)) %>

```

Issue #1433, opened 2026-05-21
ActionController::ParameterMissing when creating embeddables with no attributes


aseroff:
### Steps to Reproduce

1. Create an embeddable class that has **no attributes**:
   ```ruby
   # app/models/spina/embeds/lead_form.rb
   class Spina::Embeds::LeadForm < Spina::Embed
     # No attributes defined
   end
   ```

2. Register it in your theme's embeddables.

3. In the Trix editor, add your embed.

### Expected Behavior
The embed should be created successfully and inserted into the editor as a Trix attachment.

### Actual Behavior
The request fails with a `400 Bad Request`:

```ruby
ActionController::ParameterMissing (param is missing or the value is empty or invalid: embeddable)
```

### Root Cause

In `Spina::Admin::EmbedsController#create`:

```ruby
def embed_params
  params.require(:embeddable).permit!
end
```

When the embed has no form fields, Rails does **not** send the `embeddable: {}` parameter at all.

PR #1434, opened 2026-05-21
Allow embeddables to be blank

aseroff:
to support embeddables without any attributes.

### Context

https://github.com/SpinaCMS/Spina/issues/1433

loosen the strong params with `params.permit(embeddable: {}).fetch(:embeddable, {})`

### Guidance to review

Million ways to skin this cat, this was just my take on it.

### DIFF
--- a/app/controllers/spina/admin/embeds_controller.rb
+++ b/app/controllers/spina/admin/embeds_controller.rb
@@ -27,7 +27,7 @@ module Spina
       end
 
       def embed_params
-        params.permit(:embeddable).fetch(:embeddable, {})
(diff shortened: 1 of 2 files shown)


Bramjetten:
Looks fine, but two remarks:
- test doesn't pass
- what kind of objects without attributes do you want to embed?

aseroff:
Ah, my bad. I'll resubmit once I've figured out the test issues.

I have a form that has to sit inside text content. Previously I just set it up with an unnecessary attribute just to save it. Seemed worth fixing.

PR #1443, opened 2026-07-06
Keep the URL of live pages stable when the title changes

jeroenson:
By default the slug of a page follows its title as long as url_title is empty. That is great while building a site, but on a live page it means an innocent title edit quietly moves the page to a new URL and leaves a 301 behind. Our content team hits this every time they freshen up titles (changing a year number for example), and over time that builds redirect chains and costs link equity.

With this change the slug of a live page is locked on its next save: the current slug is written to url_title, so title edits no longer move the page. The URL can still be changed deliberately by editing url_title, which creates the usual 301. Draft pages keep following the title.

This does change behavior for existing sites, so it deserves a line in the release notes.


Bramjetten:
I don't like it as an option

jeroenson:
Fair enough. I removed the option and made it the default. Drafts keep following the title, so the only change is that a live page no longer moves to a new URL when someone edits its title. Changing the URL through url_title works like before, including the 301. The whole test suite passes with the new default. If you know a case where you actually want a title edit to silently move a live page, I'd like to hear it.

